Issue #3, opened 2019-12-12
Problem with CORS and Trinity


HarmvandenBrink:
> OPTIONS method not allowed
Preflight requests are not allowed on the node. Check your proxy configuration to allow at least POST and OPTIONS request methods.

`< HTTP/1.1 404 Not Found
< Date: Thu, 12 Dec 2019 09:47:54 GMT
< Server: Apache/2.4.29 (Ubuntu)
< Content-Type: application/json; charset=utf-8
< Content-Length: 22
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: POST, GET, OPTIONS, DELETE, PUT`

I'm using Apache as a proxy for SSL to Hornet.

PR #4, opened 2019-12-12
Fix webAPI CORS

IoTMOD:
With this fix you should be able to use HORNET nodes within Trinity (together with a reverse proxy like nginx)


IoTMOD:
Related to #3

PR #73, opened 2019-12-18
add command name to api response

Thoralf-M:
Before: 
`Request error: 'command' is protected`
`"error":"'command' parameter has not been specified"`
With this change: 
`Request error: 'attachtotangle' is protected`
`"error":"'getnodeinfoa' parameter has not been specified"`

### DIFF
--- a/plugins/webapi/api.go
+++ b/plugins/webapi/api.go
@@ -27,7 +27,7 @@ func webAPIRoute() {
 		_, permited := permitedEndpoints[cmd]
 		if apiCallExists && !permited && c.Request.RemoteAddr[:9] != "127.0.0.1" {
 			e := ErrorReturn{
-				Error: fmt.Sprintf("%v is protected", cmd),
+				Error: fmt.Sprintf("'%v' is protected", cmd),
 			}
 			c.JSON(http.StatusForbidden, e)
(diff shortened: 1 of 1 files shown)


IoTMOD:
Thanks for your PR,
could you make some adjustments?
  
The newest IRI returns this error messages:
 ```
"error": "Command [getNodeinfo] is unknown", 
```
```
"error": "COMMAND getNeighbors is not available on this node",
```

So the HORNET API errors could look like this:
```
"error": "Command [getNodeinfo] is unknown", 
```
and
```
"error": "Command [getNeighbors] is protected", 
```

And maybe we can use the command before doing `strings.ToLower()`
Something like this:
```go
/// Get the command and check if it's implemented
originCommand := fmt.Sprint(request["command"])
cmd := strings.ToLower(originCommand)

implementation, apiCallExists := implementedAPIcalls[cmd]
.
.
.
Error: fmt.Sprintf("Command [%v] is protected", originCommand),
```

Thoralf-M:
@IoTMOD thanks, updated

muXxer (approved):

muXxer:
@Thoralf-M thank you for the PR, keep up the good work!

Issue #60, opened 2019-12-17
Create an optimized DietPi image with Hornet


luca-moser:
Requirements:
* Create a DietPi image with Hornet
* The Hornet installation should be easily upgradable

bernardoaraujor:
these might help:
https://github.com/bernardoaraujor/meta-iota/wiki/HoneyComb-Distro

https://github.com/bernardoaraujor/meta-iota/wiki/BitBaking-HoneyComb-for-the-Raspberry-Pi

luca-moser:
@bernardoaraujor @Brootux We would like to split the bounty between you two. Would you guys be so kind and tell us whether this is ok and then make out which of you is going to receive the funds and then send the other half to the other?

Brootux:
sounds good for me, im also okay with "receiving as the second person"

bernardoaraujor:
@Brootux how can I get in touch with you?

Brootux:
@bernardoaraujor Should i simply post my receive address here? Does it need any more discussion? ;)

luca-moser:
release bounty to @bernardoaraujor

bernardoaraujor:
FQYGIJYKHZTLMTUMZKJKHKMWBCQTKDCGPWCRBYAAPLZRULNNAOFXQZCYQCAGUZNZLEAMTPIZQSYGGZRXAJKVCGLOJW

Brootux:
DIKVKAUGZNTKTVJFM9HAWDUYZBYFOUATRBCWLSNCQRBYMQWZCKKCLLJFAIZNEYHEZFXCWJRRMHRPRYWFXBXGQCHJLW

bernardoaraujor:
@Brootux bundle O9XHDXVRAOXEPIYZJWXBEKHCKAJBU9FWWLQJJIJJVVEYJVJYNZZHOLYEJDPAFCSHBAAWXKWGSJONRZBJA

Brootux:
Transaction received, you are an honorable man @bernardoaraujor 

Thanks everyone, this is the future, create more bounty issues ;)

PR #76, opened 2019-12-18
Brootux hornet deb packaging

Brootux:
Part of #60 

Rudimentary changes to build deb packages via goreleaser.

WIP and untested, installation should be possible but i could not test fully because systemd does not work in docker and i have currently no dedicated machine to test this.

Temporarily added script "manual_build.sh" to allow local building of deb packages in git-dirty environment and without uploading results. This script should be removed before merge.

RFC


deb_files/download_snapshot.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+if [ ! -f latest-export.gz.bin ]; then
+	wget https://dbfiles.iota.org/mainnet/hornet/latest-export.gz.bin
+fi
```
Brootux: Is this how we should handle this? Or should we download the latest-export on every service start?

manual_build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+GORELEASER_IMAGE=iotmod/goreleaser-cgo-cross-compiler:1.13.5
+REPO_PATH="/build"
+PWD="$(pwd)"
```
Brootux: only usable if executed inside the root-folder of the repository

deb_files/hornet.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Hornet
+
+[Service]
+WorkingDirectory=/var/run/hornet
+Restart=always
+ExecStartPre=/usr/bin/hornet_download_snapshot.sh
+ExecStart=/usr/bin/hornet -c /etc/hornet.conf
```
muXxer: isn't it `/etc/gohornet.conf` ?

svenger87: Should be /etc/gohornet/gohornet.conf 
Also Timeoutsec=Infinity should be set to let hornet gracefully shutdown.

manual_build.sh
```diff
@@ -0,0 +1,9 @@
...
+
+mkdir -p ${PWD}
+docker pull ${GORELEASER_IMAGE}
+docker run --rm --privileged -v $PWD:$REPO_PATH -w $REPO_PATH ${GORELEASER_IMAGE} goreleaser --release-notes=CHANGELOG.md --rm-dist --snapshot
```
IoTMOD: I guess this script is just there to check, if GoReleaser is working correctly?

* We should not publish such a manual build
* The changelog flag is outdated since we don't publish a changelog automatically anymore

```suggestion
docker run --rm --privileged -v $PWD:$REPO_PATH -w $REPO_PATH ${GORELEASER_IMAGE} goreleaser --rm-dist --snapshot --skip-publish
```

Brootux (commented):
Different questions regarding the deb-package:

- How should we handle the `/etc/hornet/hornet.conf` file? it probably gets changed with different default values in the future and needs to replaced or extended. We could place the default conf as `hornet.conf.example` and the user must copy it to `hornet.conf` with adjusted neighbors etc
- I will build it now so that after initial install, the user has to put the service into autostart by hand and also has to start it initially by hand (`systemctl enable hornet` and `systemctl start hornet`). So a user can adjust the hornet-conf first before starting the service. How should we handle the "upgrade" path?
- I would write a very short "getting started" guide, where should i place this? in the readme.md?
- Should hornet run as root or as a special user?

deb_files/download_snapshot.sh
muXxer: I was thinking about a "run.sh" script, which asks the user if he wants to delete the mainnetdb. If so, it checks automatically if the latest snapshot is still the latest. I have some codesnippets for you if you want to do that? :)

`run.sh`
```sh
#!/bin/bash
service hornet stop

echo "Do you want to delete the old database?"
select yn1 in "Yes" "No"; do
    case $yn1 in
        Yes ) rm -Rf mainnetdb/; ./download_snapshot.sh; break;;
         No ) break;;
    esac
done

service hornet start
journalctl -u hornet -f
```

`download_snapshot.sh`
```sh
#!/bin/bash

finalurl() { curl --silent --location --head --output /dev/null --write-out '%{url_effective}' -- "$@"; }
checkIfModified() { curl --silent --output /dev/null --write-out '%{http_code}' -I --header 'If-Modified-Since: Tue, 17 Dec 2019 11:56:38 GMT' -- "$@"; }

lastest_file=`finalurl https://dbfiles.iota.org/mainnet/hornet/latest-export.gz.bin`
status_code=`checkIfModified ${lastest_file}`
if [[ "$status_code" -ne 304 ]] ; then
    echo "Snapshot outdated, downloading snapshot..."
    rm latest-export.gz.bin
    wget ${lastest_file}
else
    echo "Snapshot up to date"
fi
```

Brootux: When we use systemd as init-system, i think there will be no interactive part, it only starts a service and prints stdout into "syslog" where you can see it i.e. via `journalctl -fu hornet`

Brootux: Nice snippet for outdated snapshot, i will put this in at least

Brootux: Meanwhile there is "wget -N" which activates timestamping and only downloads if remote file is newer.

Brootux (commented):
Cleaned up and tested on AMD64, feel free to test it on your own via `manual_build.sh`

rundirectory is `/var/lib/hornet` where also latest-export gets stored. On `apt purge hornet` this is currently the only directory which does not get deleted. If desired we could add cleanup to postrm-purge

deb_files/hornet.service
```diff
@@ -0,0 +1,12 @@
...
+
+[Service]
+StateDirectory=hornet
+WorkingDirectory=/var/lib/hornet
+Restart=always
+Timeoutsec=Infinity
+ExecStart=/bin/bash -c "/usr/bin/hornet_download_snapshot.sh && /usr/bin/hornet -c /etc/hornet/hornet.conf"
+
+[Install]
+WantedBy=multi-user.target
```
nuriel77: I had some ideas for the service file.
Use environment file makes it easy to configure some values (no need to issue `systemctl daemon-reload` on each change, as you would to the service file)
```
[Unit]
Description=Hornet
Wants=network-online.target
After=network-online.target

[Service]
EnvironmentFile=-/etc/default/hornet
# Use a variable from the environment file, e.g:
TimeoutStopSec=${TIMEOUT_STOP_SEC}
```
Maybe instead of:
```
ExecStart=/bin/bash -c "/usr/bin/hornet_download_snapshot.sh && /usr/bin/hornet -c /etc/hornet/hornet.conf"`
```
do:
```
ExecStartPre=-/bin/bash /usr/bin/hornet_download_snapshot.sh
ExecStart=/path/to/bin/hornet $OPTIONS # options can be configured in the environment file, should the user want to add
```

Perhaps add (and pre-create the user group so not to run as root):
```
User=hornet
Group=hornet
WorkingDirectory=/path/to/hornet/workdir
```

Add security, logging, optional open files limits:
```
LimitNOFILE=4096
StandardOutput=syslog
StandardError=syslog
SyslogIdentifier=hornet

PrivateDevices=yes
PrivateTmp=yes
ProtectSystem=full
ProtectHome=yes
```

deb_files/postinst
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -e
```
nuriel77, line 3: You've used `|| true` on most commands, then why `set -e`?

deb_files/postinst
```diff
@@ -0,0 +1,20 @@
...
+SERVICE_NAME="hornet"
+SYSTEMD_SERVICE="${SERVICE_NAME}.service"
+
+case "$1" in
+    configure|abort-upgrade|abort-remove|abort-deconfigure)
+        deb-systemd-helper unmask "${SYSTEMD_SERVICE}" >/dev/null || true
+        systemctl --system daemon-reload >/dev/null || true
+        if [ -n "$2" ]; then
+            deb-systemd-invoke restart "${SYSTEMD_SERVICE}" >/dev/null || true
+	fi
```
nuriel77: indentation +2 spaces

deb_files/postrm
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+
+set -e
```
nuriel77, line 3: Same, why use `set -e` if defaulting all commands to `|| true` if they fail?
Would make sense to remove `|| true` and let `set -e` do its job, or add proper error handling for commands.

manual_build.sh
nuriel77: Isn't `$PWD` already set to the same value as `$(pwd)`?
```
root@ubuntu-4gb-fsn1-1:~# pwd
/root
root@ubuntu-4gb-fsn1-1:~# echo $PWD
/root
```

manual_build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+GORELEASER_IMAGE=iotmod/goreleaser-cgo-cross-compiler:1.13.5
+REPO_PATH="/build"
+PWD="$(pwd)"
+
+mkdir -p ${PWD}
```
nuriel77: Always good to use double quotes on variables in bash unless explicitly not required (e.g. multiple cli options provided variable)

Brootux:
While testing the package i found out that every time my node was synced and then restarted, i got these "please reset the database" error, is this currently normal?

IoTMOD:
This should only happen if HORNET has received a kill signal (by systemd etc.) because the shutdown took too long. The shutdown may take a few minutes depending on the hardware.
Or there was a OOM while shutting down.

Brootux:
- Rebased to master
- Found the problem regarding the service restart
- Adjusted for the new confdir/conf-string/neighbors-string parameter interface
- Fixed bugs regarding the helper-script

Removing WIP since i would consider current state as ready

.goreleaser.yml
```diff
@@ -130,6 +130,35 @@ archives:
       - config.json
       - neighbors.json
 
+# Additional packages (i.e. deb)
+nfpms:
+-
+  id: nfpm
+  package_name: hornet
+  file_name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
```
IoTMOD: We should use the dpkg name convention:
https://manpages.debian.org/buster/dpkg-dev/dpkg-name.1.en.html

```suggestion
  file_name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Arch }}"
```

scripts/build_goreleaser_snapshot.sh
```diff
@@ -0,0 +1,20 @@
...
+SCRIPTDIR=$(dirname "$0")
+pushd "${SCRIPTDIR}" >/dev/null
+pushd ".." >/dev/null
+
+
+GORELEASER_IMAGE=iotmod/goreleaser-cgo-cross-compiler:1.13.5
+REPO_PATH="/build"
+
+docker pull "${GORELEASER_IMAGE}"
+docker run --rm --privileged -v "${PWD}":"${REPO_PATH}" -w "${REPO_PATH}" "${GORELEASER_IMAGE}" goreleaser --rm-dist --snapshot
```
IoTMOD: Please add a `--skip-publish` (e.g. to not publish a docker image by mistake, if you are logged in into the docker hub)
```suggestion
docker run --rm --privileged -v "${PWD}":"${REPO_PATH}" -w "${REPO_PATH}" "${GORELEASER_IMAGE}" goreleaser --rm-dist --snapshot --skip-publish
```

IoTMOD (commented):
Hi, 
that's looking good :+1:
In addition to the two suggestions, please change the PR target to the `develop` branch

Brootux:
@IoTMOD fixed all your recommendations and targeting `develop` branch now

IoTMOD (approved):

IoTMOD:
Thank you very much!

PR #88, opened 2019-12-20
Add arm64 Dockerfile, add timeout on gpg receive key

nuriel77:
# Dockerfile for arm64

* Adds a Dockerfile specific for ARM64.
I didn't add ARM32 support at this time, as I think that with 32bit a user best use the binaries and skip running hornet in Docker.

* Add timeout on gpg key receive
Some gpg servers in the list may timeout, gpg doesn't have any flag to specify the timeout and it takes quite long. Using the `timeout` (which is available on alpine by default) helps terminate the process so that the next server in the list is attempted.

Note: It would have been ideal to reuse the original Dockerfile in the repository, however there are two ARGs that would have to be passed which makes it confusing for the user.
Idea was to set `--build-arg arm64` and set `tini`'s version accordingly. However, the image name requires the prefix `arm64v8/`, that would require to pass that too as an argument. There's not much that can be done with the `ARG` keyword in Dockerfile as they only support very limited parameter expansion.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -41,6 +41,11 @@ If not running via docker-compose, build manually:
 docker build -t hornet:latest .
 ```
 
+Note: for aarch64/arm64 architecture point docker build to the `Dockerfile.arm64`:
+```sh
+docker build -f Dockerfile.arm64 -t hornet:latest .
+```
+
 ## Run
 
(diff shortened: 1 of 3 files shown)


IoTMOD:
Hi @nuriel77,
could you add a short note to the Docker.md pointing out there is an arm64 Dockerfile now?

IoTMOD (approved):

PR #130, opened 2020-01-09
feat: Add neighbor config observer

IoTMOD:
- [x] We need to decide whether a non-resolvable neighbor should be added at all


plugins/gossip/neighbor_config_observer.go
```diff
@@ -0,0 +1,146 @@
...
+			modifyNeighbors(modified)
+		}
+
+		// Add neighbors
+		if len(added) > 0 {
+			addNewNeighbors(added)
+		}
+
+		// Remove Neighbors
+		// FixMe: Removing example neighbors will result in an error, but the neighbour will still be deleted.
```
golangcibot: `neighbour` is a misspelling of `neighbor` (from `misspell`)
```suggestion

```

alexsporn (approved):

PR #131, opened 2020-01-11
RPM Packages

nuriel77:
# Overview
* adds RPM package build
* move directory structure
* minor stuff like scripts polishing

# Test
Tested by running yum install the package (CentOS 7).
All files are where they are supposed to be.
Hornet starts up.
Tested yum remove.


scripts/build_goreleaser_snapshot.sh
```diff
@@ -4,17 +4,11 @@
 # Allows building of dirty git state and will not push results to github.
 
 # make script executable independent of path
-SCRIPTDIR=$(dirname "$0")
-pushd "${SCRIPTDIR}" >/dev/null
-pushd ".." >/dev/null
-
+trap 'cd $(dirs -l -0)' EXIT INT QUIT TERM
```
nuriel77: Less lines of codes to achieve the same as the pushd/popd.

IoTMOD (approved):

IoTMOD:
Thank you very much! :+1:

PR #142, opened 2020-01-15
allow other orders of txs in attachToTangle and validate indexes

Thoralf-M:
Issue for IRI: https://github.com/iotaledger/iri/issues/1708
With this change the txs doesn't have to be send in the correct order to attachToTangle to get a valid bundle and PoW will not be done for bundles which would have invalid consistency


luca-moser (dismissed):
Instead of making a new slice for when sorting a transaction, simply reorder the transactions within the origin request slice in place, if you see that the first transaction in the slice is the tail (currentIndex 0).

IoTMOD:
You could use something like this:
```go
// Sort transactions (highest to lowest index)
sort.Slice(txs, func(i, j int) bool {
	return txs[i].CurrentIndex > txs[j].CurrentIndex
})
```

And please do this PR against the `develop` branch according to https://github.com/gohornet/hornet/blob/master/CONTRIBUTING.md

IoTMOD (approved):

IoTMOD:
Thanks for this :+1:

PR #149, opened 2020-01-18
- Moving storing of new incoming tx into the bundle bucket

alexsporn:
Incoming Tx are now stored to the OS in AddTransaction(), so that the bucket is locked when a new tx is added to the storage

This way we make sure that no solidifier can try to solidify this tx before it was processed in the bundle bucket

### DIFF
--- a/packages/model/tangle/bundle.go
+++ b/packages/model/tangle/bundle.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/iotaledger/hive.go/bitmask"
 	"github.com/iotaledger/hive.go/syncutils"
+
+	"github.com/gohornet/hornet/packages/model/hornet"
 )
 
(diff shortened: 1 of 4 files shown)


packages/model/tangle/bundle.go
```diff
@@ -229,21 +231,28 @@ func (bucket *BundleBucket) GetHash() trinary.Hash {
...
 
+	tx := GetCachedTransaction(hornetTx.GetHash())
+	defer tx.Release()
+
+	if tx.Exists() {
+		return nil, true
+	}
+
+	// Store the tx in the storage, this will update the tx reference automatically
+	StoreTransaction(hornetTx).Release()
```
muXxer, line 247: Isn't `Release` already called by defer?

muXxer (approved):

PR #174, opened 2020-01-25
Update deb postinst

miili:
The service defines `WorkingDirectory=/var/lib/hornet`, which is now created.


IoTMOD:
Hi @miili,
thanks for your PR.

It's already fixed with #141 
The fix will be included within the next HORNET release

PR #179, opened 2020-01-27
Improve the RPM/DEB packages, alternative d/l option

nuriel77:
# Motivation
I'd like to specify an alternative location where to get the snapshot file from.
Or, I'd like to specify some additional command line options for hornet (eventhough the cli options exist in the config file, it is sometimes quicker/handier to add something like `--node.LogLevel=127 --useProfile=1gb` in the env file and restart hornet)

# Changes
* Add load of environment file to help configure options (e.g. which remote snapshot file URL to use or cli options). Such file is quite common on services installed on linux to enter extra command line options etc.

# Minor Fixes
* Don't recursively chmod 700 directories as that makes any files in there executable.
* Sudo missing after a &&
* Make sure we enter the repository's root. There seems to be no need to cd back to original dir: when script ends, the dir is the original one.

### DIFF
--- a/nfpm/shared_files/hornet.env
+++ b/nfpm/shared_files/hornet.env
@@ -1,0 +1,5 @@
+# Use an alternative snapshot file location
+#REMOTE_FILE="http://my-alternative.location.io/file.bin.gz"
+
+# Add cli arguments to hornet
+#OPTIONS="--node.LogLevel=127 --useProfile=1gb"
--- a/nfpm/shared_files/hornet.env
+++ b/nfpm/shared_files/hornet.env
@@ -1,5 +1,6 @@
-# Use an alternative snapshot file location
+# Use an alternative snapshot file location, e.g.:
 #REMOTE_FILE="http://my-alternative.location.io/file.bin.gz"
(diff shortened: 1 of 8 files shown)


nuriel77:
Tested the RPM and DEB packages install & remove.

nfpm/shared_files/hornet.env
```diff
@@ -0,0 +1,5 @@
+# Use an alternative snapshot file location
+#REMOTE_FILE="http://my-alternative.location.io/file.bin.gz"
+
+# Add cli arguments to hornet
+#OPTIONS="--node.LogLevel=127 --useProfile=1gb"
```
IoTMOD: The `--node.LogLevel=127` cli arg got removed from HORNET (with HORNET v0.3.0).
Please add an alternative example (available cli args: `./hornet -h`) or remove it.

IoTMOD (changes requested):

IoTMOD (approved):
Thanks for this! 
We really appreciate your contribution to HORNET :+1:

PR #187, opened 2020-02-05
Adds permaspent plugin

luca-moser:
* Adds a plugin which forwards `wereAddressesSpentFrom` to multiple [permaspent](https://github.com/gohornet/permaspent) nodes, with config keys for quorum and error thresholds.
* Local snapshot code doesn't read or write spent addresses anymore. Such files are automatically flagged as version `4`, while also files with version `3` can be imported.
* The SPA no longer shows any information about spent addresses.
* `markedSpentAddr` has been renamed to `seenSpentAddr`
* `getNodeInfo` returns `WereAddressesSpentFrom` in the `features` set if the `permaspent` plugin is enabled.
* Cuckoo filter and other spent addresses related code has been removed
* Added `spent_address` topic to ZMQ and MQTT plugins


IoTMOD (approved):

main.go
```diff
@@ -5,6 +5,7 @@ import (
 	_ "net/http/pprof"
 	"runtime"
 
+	"github.com/gohornet/hornet/plugins/permaspent"
```
muXxer: :rage:

plugins/mqtt/plugin.go
```diff
@@ -38,6 +39,10 @@ var (
 	newSolidMilestoneWorkerQueueSize = 100
 	newSolidMilestoneWorkerPool      *workerpool.WorkerPool
 
+	spentAddressWorkerCount     = 1
+	spentAddressWorkerQueueSize = 100
```
muXxer: Better 1000 here while syncing?

muXxer (approved):

PR #188, opened 2020-02-10
Update Snapshot URL

centercirclesolutions:
Old URL is no longer up to date. Use https://ls.manapotion.io/latest-export.gz.bin or until a better option hosted by IF is available


muXxer:
IF updated the snapshots.... we will let this PR open some days to check if the IF snapshots stay up to date

centercirclesolutions:
@muXxer @luca-moser IF site is looking good. I will close this later today unless I hear otherwise

PR #203, opened 2020-02-29
Added support for querying the external IP address and opening port mappings using NAT-PMP

alexsporn:
Use  network.externalAddress == "nat-pmp" instead of "auto"


alexsporn:
This will need too much support, so let the users open their ports manually

PR #211, opened 2020-03-03
Reintroduce spent addr

muXxer:
This PR re-adds the spent addresses in the database.

Now the user can choose if he wants to have spent addresses with a config flag.
The API call wereAddressesSpentFrom is blocked if spent addresses were not loaded at startup.


packages/model/tangle/spent_addresses_storage.go
```diff
@@ -0,0 +1,170 @@
...
+		select {
+		case <-abortSignal:
+			wasAborted = true
+			return false
+		default:
+		}
+
+		addressesWritten++
+		err = binary.Write(buf, binary.LittleEndian, key)
+		if err != nil {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

luca-moser (approved):

PR #212, opened 2020-03-04
Update file to download snapshot

centercirclesolutions:
Inline with changes by Luca


muXxer:
Thx Dave for the PR, unfortunately the official URL will be different :(
We will change it as soon as snapshots use the new address/name.

PR #224, opened 2020-03-08
Add API whitelist

viossat:
# Description

Allow specified addresses in `config.json` to access all API commands.
Accepts both IP and CIDR. `127.0.0.1` and `::1` are hard-coded as whitelisted.

To accept all IP:
```json
{
  "api": {
    "whitelistedAddresses": ["0.0.0.0/0", "::/0"]
  }
}
```

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

# How Has This Been Tested?

I tried to access permitted and protected API commands from whitelisted and unwhitelisted addresses.

# Checklist:

### DIFF
--- a/config.json
+++ b/config.json
@@ -18,6 +18,7 @@
       "storeTransactions",
       "getTrytes"
     ],
+    "whitelistedAddresses": [],
     "bindAddress": "0.0.0.0",
     "maxbodylength": 1000000,
     "maxfindtransactions": 1000,
--- a/plugins/webapi/plugin.go
+++ b/plugins/webapi/plugin.go
@@ -3,12 +3,14 @@ package webapi
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/gin-contrib/gzip"
 	"github.com/gin-gonic/gin"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
 
 	"github.com/iotaledger/hive.go/daemon"
 	"github.com/iotaledger/hive.go/logger"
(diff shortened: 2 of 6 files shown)


config.json
```diff
@@ -18,6 +18,7 @@
       "storeTransactions",
       "getTrytes"
     ],
+    "whitelistedAddresses": [],
```
nuriel77: Afaik, current behavior is to whitelist `::1` and `127.0.0.1` by default, maybe this should be specified here by default, keeps it more visible, avoid confusion?

plugins/webapi/plugin.go
```diff
@@ -75,6 +78,19 @@ func configure(plugin *node.Plugin) {
 		}
 	}
 
+	// Load whitelisted addresses
+	whitelist := append([]string{"127.0.0.1", "::1"}, parameter.NodeConfig.GetStringSlice("api.whitelistedAddresses")...)
+	if len(whitelist) > 0 {
```
IoTMOD: `len(whitelist)` is always 2 here, so you don't need this check

plugins/webapi/plugin.go
```diff
@@ -75,6 +78,19 @@ func configure(plugin *node.Plugin) {
...
+		for _, entry := range whitelist {
+			_, ipnet, err := cnet.ParseCIDROrIP(entry)
+			if err != nil {
+				log.Warnf("Invalid whitelist address: %s", entry)
+			} else {
+				whitelistedNetworks = append(whitelistedNetworks, ipnet.IPNet)
+			}
```
IoTMOD: We can remove the `else` here
```suggestion
			if err != nil {
				log.Warnf("Invalid whitelist address: %s", entry)
				continue
			} 
		        whitelistedNetworks = append(whitelistedNetworks, ipnet.IPNet)
```

IoTMOD (changes requested):
Thanks for your PR!

plugins/webapi/plugin.go
viossat: True, I added the two addresses after the condition

IoTMOD (approved):

PR #238, opened 2020-03-11
Validate strict bundle semantics up on solidification too

luca-moser:
This PR adds a new metadata flag to a bundle instance telling whether a bundle is valid regarding strict bundle semantics:
* the non head txs of a bundle approve through their branch the trunk transaction of the head tx
* the head tx only approves tail transactions

The strict bundle semantics are checked during tip-selection, respectively when computing a cone diff of a given trasaction.

This PR enforces that non milestone bundles are solid during solidification as the 2nd check needs that. Milestone bundles are of course still validated and it is assumed that they obey to the strict bundle semantics.

### DIFF
--- a/packages/model/tangle/bundle.go
+++ b/packages/model/tangle/bundle.go
@@ -238,7 +238,23 @@ func (bundle *Bundle) validate(onMaybeMilestone func() bool) bool {
 		cachedCurrentTx.Release(true) // tx -1
 	}
 
+
+	// validate bundle semantics and signatures
+	if iotago_bundle.ValidBundle(iotaGoBundle) != nil {
+		bundle.setValid(false)
+		bundle.setValidStrictSemantics(false)
+		return false
+	}
+
 	validStrictSemantics := true
+
+	// verify that the head transaction only approves tail transactions.
+	// this is fine within the validation code, since the bundle is only complete when it is solid.
+	// however, as a special rule, milestone bundles might not be solid
+	checkTails := true
+	if IsMaybeMilestone(cachedCurrentTailTx) {
+		checkTails = !onMaybeMilestone()
+	}
 
 	// enforce that non head transactions within the bundle approve as their branch transaction
 	// the trunk transaction of the head transaction.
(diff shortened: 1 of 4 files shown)


packages/model/tangle/bundle.go
```diff
@@ -215,22 +226,74 @@ func (bundle *Bundle) validate() bool {
...
+	defer cachedCurrentTailTx.Release(true) // tx -1
 
+	cachedCurrentTx := cachedCurrentTailTx
 	for i := 1; i < lastIndex+1; i++ {
 		cachedCurrentTx = loadBundleTxIfExistsOrPanic(cachedCurrentTx.GetTransaction().GetTrunk(), bundle.hash) // tx +1
 		iotaGoBundle[i] = *cachedCurrentTx.GetTransaction().Tx
 		cachedCurrentTx.Release(true) // tx -1
 	}
 
+	validStrictSemantics := true
```
muXxer: Why not move that check before the `if checkTails {`

plugins/tangle/consistency.go
```diff
@@ -3,6 +3,7 @@ package tangle
 import (
 	"errors"
 
+	"github.com/gohornet/hornet/packages/model/hornet"
```
muXxer: move that to the other gohornet import pls

muXxer (changes requested):

packages/model/tangle/bundle.go
```diff
@@ -215,22 +226,75 @@ func (bundle *Bundle) validate() bool {
...
+	defer cachedCurrentTailTx.Release(true) // tx -1
 
+	cachedCurrentTx := cachedCurrentTailTx
 	for i := 1; i < lastIndex+1; i++ {
 		cachedCurrentTx = loadBundleTxIfExistsOrPanic(cachedCurrentTx.GetTransaction().GetTrunk(), bundle.hash) // tx +1
 		iotaGoBundle[i] = *cachedCurrentTx.GetTransaction().Tx
 		cachedCurrentTx.Release(true) // tx -1
 	}
 
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

muXxer (approved):

PR #245, opened 2020-03-12
Fix local snapshots if txs reference non-tails

muXxer:
This PR fixes the local snapshot creation, if the cone that gets approved by a milestone contains transactions that reference non-tail transactions of another bundle.

### DIFF
--- a/plugins/snapshot/local_snapshot.go
+++ b/plugins/snapshot/local_snapshot.go
@@ -122,7 +122,9 @@ func getMilestoneApprovees(milestoneIndex milestone_index.MilestoneIndex, cached
 						log.Panicf("getMilestoneApprovees: Transaction not confirmed: %v", txHash)
 					}
 
-					// Search all referenced tails of this Tx (needed for correct SolidEntryPoint calculation)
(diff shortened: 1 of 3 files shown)


plugins/snapshot/local_snapshot.go
```diff
@@ -116,20 +116,41 @@ func getMilestoneApprovees(milestoneIndex milestone_index.MilestoneIndex, cached
...
+					}
+
+					for tailTx := range tailTxs {
+						txsToTraverse[tailTx] = struct{}{}
+					}
+
+					// Ignore this transaction in the cone because it is not confirmed
 					cachedTx.Release(true) // tx -1
 					continue
+				} else {
```
luca-moser: This else statement is redundant since the if always continues the loop or returns out of the function

luca-moser (changes requested):

plugins/snapshot/local_snapshot.go
muXxer: Good catch

luca-moser (approved):

PR #252, opened 2020-03-13
Refactors configuration options

luca-moser:
* Normalizes config keys to snakeCase
* Use variables instead of plain strings to access config options
* Renamed `parameters` package to `config`
* Removed all `parameters.go` files from plugins in favor of having files within the `config` package
* HTTP basic auth middleware for the dashboard and web API now use sha256 hashes with a given salt instead of having passwords in plain in the `config.json`
* Added an own implementation of a HTTP basic auth middleware for the web API as Gin doesn't provide an actual middleware where one can inspect the supplied username/password combination, which makes it impossible to hash those values to compare them against the config values
* Merged autopeering, gossping and networking options into one JSON object
* Merged local and global snapshot options into one JSON object
* Defining a snapshot type to load which isn't `local` or `global` lets the node exit with an error message
* All `bindAddress` options now include the port too. It makes sense to have the address and port as one string (as that is the default Go way of having such construct), however in the case of the gossipping server, the `bindAddress` still has to be deconstructed as the `TCPServer` in `hive.go` uses two parameters instead of simply taking a string. (probably something to change in `hive.go`)

It is suggested that node owners simply use the new config.json and transfer values from their current over to the new one, as a lot of keys have changed (instead of mutating their current one).


packages/config/zmq_config.go
```diff
@@ -0,0 +1,13 @@
+package config
+
+const (
+	// protocol used to connect to the zmq feed [unix, tcp, udp, inproc]
+	CfgZMQBindAddress = "zmq.bindAddress"
+	// the bind address of the ZMQ feed
+	CfgZMQProtocol    = "zmq.protocol"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	CfgZMQProtocol = "zmq.protocol"
```

packages/config/dashboard_config.go
```diff
@@ -0,0 +1,28 @@
...
+	// the HTTP basic auth password+salt as a sha256 hash
+	CfgDashboardBasicAuthPasswordHash = "dashboard.basicauth.passwordhash" // must be lower cased
+	// the HTTP basic auth salt used for hashing the password
+	CfgDashboardBasicAuthPasswordSalt = "dashboard.basicauth.passwordsalt" // must be lower cased
+)
+
+func init() {
+	NodeConfig.SetDefault(CfgDashboardBindAddress, "localhost:8081")
+	NodeConfig.SetDefault(CfgDashboardDevMode, false)
+	NodeConfig.SetDefault(CfgDashboardBasicAuthEnabled, true)
```
muXxer: The default should be "false"

packages/config/network_config.go
```diff
@@ -0,0 +1,62 @@
...
+	NodeConfig.SetDefault(CfgNetPreferIPv6, false)
+	NodeConfig.SetDefault(CfgNetGossipBindAddress, "0.0.0.0")
+	NodeConfig.SetDefault(CfgNetGossipReconnectAttemptIntervalSeconds, 60)
+
+	// neighbors
+	NeighborsConfig.SetDefault(CfgNeighborsAcceptAnyNeighborConnection, false)
+	NeighborsConfig.SetDefault(CfgNeighborsMaxNeighbors, 5)
+	NeighborsConfig.SetDefault(CfgNeighbors, []NeighborConfig{
+		{
+			Identity:   "example1.neighbor.com:15600",
```
muXxer: Please use "example.neighbor.com:15600" because that is ignored in `AddNeighbor`

packages/config/tipsel_config.go
```diff
@@ -0,0 +1,14 @@
...
+const (
+	// the max allowed depth to be used as the starting point for tip-selection
+	CfgTipSelMaxDepth = "tipsel.maxDepth"
+	// the limit defining the max amount of transactions to traverse in order to check
+	// whether a transaction references a transaction below max depth
+	CfgTipSelBelowMaxDepthTransactionLimit = "tipsel.belowMaxDepthTransactionLimit"
+)
+
+func init() {
+	NodeConfig.SetDefault(CfgTipSelMaxDepth, 15)
```
muXxer: `3` should be default.

packages/config/webapi_config.go
```diff
@@ -0,0 +1,53 @@
...
+			"getInclusionStates",
+			"getNodeAPIConfiguration",
+			"wereAddressesSpentFrom",
+			"broadcastTransactions",
+			"findTransactions",
+			"storeTransactions",
+			"getTrytes",
+		})
+	NodeConfig.SetDefault(CfgWebAPIWhitelistedAddresses, []string{})
+	NodeConfig.SetDefault(CfgWebAPIBasicAuthEnabled, "")
```
muXxer: Should be `false`

packages/config/webapi_config.go
```diff
@@ -0,0 +1,53 @@
...
+	NodeConfig.SetDefault(CfgWebAPIBasicAuthUsername, "")
+	NodeConfig.SetDefault(CfgWebAPIBasicAuthPasswordHash, "")
+	NodeConfig.SetDefault(CfgWebAPIBasicAuthPasswordSalt, "")
+	NodeConfig.SetDefault(CfgWebAPILimitsMaxGetTrytes, 10000)
```
muXxer: Should be `1000`

packages/config/webapi_config.go
```diff
@@ -0,0 +1,53 @@
...
+	NodeConfig.SetDefault(CfgWebAPILimitsMaxRequestsList, 1000)
+	NodeConfig.SetDefault(CfgWebAPILimitsMaxFindTransactions, 100000)
```
muXxer: Should be `1000`

muXxer (changes requested):

packages/config/webapi_config.go
luca-moser: Changed

packages/config/webapi_config.go
luca-moser: changed

packages/config/webapi_config.go
luca-moser: changed

packages/config/tipsel_config.go
luca-moser: changed

packages/config/network_config.go
luca-moser: changed

packages/config/dashboard_config.go
luca-moser: changed

muXxer (approved):

PR #287, opened 2020-03-20
Update to new autopeering version

muXxer:
This PR updates to the latest autopeering version in hive.go.
This has the advantage that no external IP lookup service is required.


plugins/autopeering/autopeering.go
```diff
@@ -56,55 +66,43 @@ func configureAP() {
...
-	gossipHost, _, err := net.SplitHostPort(gossipAddr.String())
-	if err != nil {
-		return false
-	}
-	peeringAddr := p.Services().Get(service.PeeringKey)
-	peeringHost, _, err := net.SplitHostPort(peeringAddr.String())
-	if err != nil {
+
+	// gossip service must be valid
+	if gossipService.Network() != "tcp" || gossipService.Port() < 0 || gossipService.Port() > 65535 {
```
Wollac: Valid port numbers are 1 through 65535. Consider using `netutil.IsValidPort`

plugins/autopeering/autopeering.go
```diff
@@ -136,25 +132,21 @@ func parseEntryNodes() (result []*peer.Peer, err error) {
...
 
-		entryAddr, err := iputils.ParseOriginAddress(parts[1])
+		pubKey, err := base64.StdEncoding.DecodeString(parts[0])
 		if err != nil {
-			return nil, fmt.Errorf("%w: invalid entry node address %s", err, parts[1])
+			return nil, fmt.Errorf("%w: invalid public key: %s", ErrParsingEntryNode, err)
 		}
 
-		ipAddresses, err := iputils.GetIPAddressesFromHost(entryAddr.Addr)
+		addr, err := net.ResolveUDPAddr("udp", parts[1])
```
Wollac: This is handled a bit differently as before. Before iputils was used to get the IP address while no the `net` is used directly. Should be more or less equivalent, the only potential difference I can see is that in this version, no longer considers `config.CfgNetPreferIPv6` when the host can be resolved to ipv4 and ipv6. But I am not sure this actually makes a difference

plugins/autopeering/local/local.go
```diff
@@ -26,41 +26,55 @@ var (
...
-			log.Fatalf("Error querying external IP: %s", err)
-		}
-		log.Infof("External IP queried: address=%s", ip.String())
-		externalIP = ip
+		// let the autopeering discover the IP
+		peeringIP = net.IPv4zero
 	} else {
-		externalIP = net.ParseIP(str)
-		if externalIP == nil {
+		peeringIP = net.ParseIP(str)
```
Wollac: It might be worth considering completely removing the `config.CfgNetAutopeeringExternalAddr` option (and always assume "auto") the case where a different external IP is specified is currently not supported by the autopeering version this is using. So, this option could just lead to confusion.

plugins/autopeering/local/local.go
```diff
@@ -26,41 +26,55 @@ var (
...
 	if str := config.NodeConfig.GetString(config.CfgNetAutopeeringExternalAddr); strings.ToLower(str) == "auto" {
-		log.Info("Querying external IP ...")
-		ip, err := netutil.GetPublicIP(config.NodeConfig.GetBool(config.CfgNetPreferIPv6))
-		if err != nil {
-			log.Fatalf("Error querying external IP: %s", err)
-		}
-		log.Infof("External IP queried: address=%s", ip.String())
-		externalIP = ip
+		// let the autopeering discover the IP
+		peeringIP = net.IPv4zero
```
Wollac: It is just cosmetic, but maybe consider using `net.IPv6unspecified` when IP v6 is preferred.
But this wont have any impact on the behavior, e.g. the pong will always be send to the source IP of the ping, using IPv4 is the source is v4 and IPv6 if the source is v6.

plugins/autopeering/local/local.go
```diff
@@ -26,41 +26,55 @@ var (
...
 		}
-		gossipBindAddr := net.JoinHostPort(externalIP.String(), gossipBindAddrPort)
-		ownServices.Update(services.GossipServiceKey(), "tcp", gossipBindAddr)
+
+		gossipBindAddrPort, err := strconv.Atoi(gossipBindAddrPortStr)
+		if err != nil {
+			log.Fatalf("Invalid gossip port number: %s, Error: %s", gossipBindAddrPortStr, err)
+		}
+
+		if 0 > gossipBindAddrPort || gossipBindAddrPort > 65535 {
```
Wollac: Maybe use `netutil.IsValidPort`

plugins/autopeering/local/local.go
```diff
@@ -26,41 +26,55 @@ var (
...
+	}
+
 	// announce the peering service
 	ownServices := service.New()
-	ownServices.Update(service.PeeringKey, "udp", net.JoinHostPort(externalIP.String(), peeringPort))
+	ownServices.Update(service.PeeringKey, "udp", peeringPort)
+
 	if !config.NodeConfig.GetBool(config.CfgNetAutopeeringRunAsEntryNode) {
-		_, gossipBindAddrPort, err := net.SplitHostPort(config.NodeConfig.GetString(config.CfgNetGossipBindAddress))
+		_, gossipBindAddrPortStr, err := net.SplitHostPort(config.NodeConfig.GetString(config.CfgNetGossipBindAddress))
```
Wollac, line 52: What happens when the IP part of `config.CfgNetGossipBindAddress` is different from  the IP part of `config.CfgNetAutopeeringBindAddr`? I guess, currently nothing not even a warning. Is this expected? Maybe make only the gossip port configurable?

plugins/autopeering/local/local.go
Wollac: Join this with the `if err != nil` from line 69, as it is the same error and should produce the same error message.

plugins/gossip/autopeering.go
```diff
@@ -46,27 +48,31 @@ func configureAutopeering() {
 		if !ev.Status {
 			return // ignore rejected peering
 		}
-		gossipAddr := ev.Peer.Services().Get(services.GossipServiceKey()).String()
-		apLog.Infof("[incoming peering] whitelisting %s / %s / %s", ev.Peer.Address(), gossipAddr, ev.Peer.ID())
+		gossipAddr := ev.Peer.Services().Get(services.GossipServiceKey())
```
Wollac: technically, this is no longer an "address" but only network (tcp or udp) plus a port. So consider renaming to gossipService or something similar

plugins/gossip/autopeering.go
```diff
@@ -46,27 +48,31 @@ func configureAutopeering() {
...
+
 		// will be grabbed later by the incoming connection
-		allowedIdentities[gossipAddr] = ev.Peer
+		allowedIdentities[gossipAddrStr] = ev.Peer
+
 		// remove from host blacklist
 		hostsBlacklistLock.Lock()
-		originAddr, _ := iputils.ParseOriginAddress(gossipAddr)
-		delete(hostsBlacklist, originAddr.Addr)
+		delete(hostsBlacklist, ev.Peer.Address().IP.String())
```
Wollac: Why `ev.Peer.Address().IP` instead of `ev.Peer.IP().String()`?
Before exactly the IP from the gossip service was used, this would correspond to `ev.Peer.IP()`

plugins/gossip/autopeering.go
```diff
@@ -46,27 +48,31 @@ func configureAutopeering() {
...
 	}))
 
 	selection.Events.OutgoingPeering.Attach(events.NewClosure(func(ev *selection.PeeringEvent) {
 		if !ev.Status {
 			return // ignore rejected peering
 		}
-		gossipAddr := ev.Peer.Services().Get(services.GossipServiceKey()).String()
-		apLog.Infof("[outgoing peering] adding autopeering neighbor %s / %s / %s", ev.Peer.Address(), gossipAddr, ev.Peer.ID())
-		if err := AddNeighbor(gossipAddr, false, "", ev.Peer); err != nil {
+		gossipAddr := ev.Peer.Services().Get(services.GossipServiceKey())
```
Wollac: Rename to gossipService or similar

plugins/gossip/neighbors.go
```diff
@@ -281,8 +281,9 @@ func finalizeHandshake(protocol *protocol, handshake *Handshake) error {
 		// grab autopeering information from whitelist
 		neighbor.Autopeering = allowedIdentities[neighbor.Identity]
 		if neighbor.Autopeering != nil {
-			gossipAddr := neighbor.Autopeering.Services().Get(services.GossipServiceKey()).String()
-			gossipLogger.Infof("handshaking with autopeered neighbor %s / %s", gossipAddr, neighbor.Autopeering.ID())
+			gossipAddr := neighbor.Autopeering.Services().Get(services.GossipServiceKey())
```
Wollac: Rename to gossipService or similar

plugins/gossip/reconnect_pool.go
```diff
@@ -119,8 +119,9 @@ next:
 		neighborsLock.Unlock()
 
 		if neighbor.Autopeering != nil {
-			gossipAddr := neighbor.Autopeering.Services().Get(services.GossipServiceKey()).String()
-			gossipLogger.Infof("initiating connection to autopeered neighbor %s / %s", gossipAddr, neighbor.Autopeering.ID())
+			gossipAddr := neighbor.Autopeering.Services().Get(services.GossipServiceKey())
```
Wollac: Rename to gossipService or similar

Wollac (changes requested):

packages/config/network_config.go
```diff
@@ -2,9 +2,9 @@ package config
 
 // NeighborConfig struct
 type NeighborConfig struct {
-	Identity   string `mapstructure:"identity"`
-	Alias      string `mapstructure:"alias"`
-	PreferIPv6 bool   `mapstructure:"preferIPv6"`
+	Identity   string `json:"identity",mapstructure:"identity"`
```
golangcibot: structtag: struct field tag `json:"identity",mapstructure:"identity"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces (from `govet`)

packages/config/network_config.go
```diff
@@ -2,9 +2,9 @@ package config
...
+	Alias      string `json:"alias",mapstructure:"alias"`
```
golangcibot: structtag: struct field tag `json:"alias",mapstructure:"alias"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces (from `govet`)

packages/config/network_config.go
```diff
@@ -2,9 +2,9 @@ package config
...
+	PreferIPv6 bool   `json:"preferIPv6",mapstructure:"preferIPv6"`
```
golangcibot: structtag: struct field tag `json:"preferIPv6",mapstructure:"preferIPv6"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces (from `govet`)

luca-moser (approved):

PR #315, opened 2020-03-23
improvements to .deb

bernardoaraujor:
- enforce creation of `/var/lib/hornet/db` and `/var/lib/hornet/snapshot` on postinst
- remove unnecessary commands from `postrm`
  - `hornet` user was replaced by `beekeeper`, which should stay in the system
  - `/etc/default/hornet/`is removed by the package manager automatically, no need to enforce it.


bernardoaraujor:
additional improvements:
- replace `beekeeper` user with `hornet`
- fix bug in `ExecStart` line inside `hornet.service`
- revert `db` and `snapshot` subdirectories (keep everything unified inside `/var/lib/hornet`)

IoTMOD (approved):

PR #318, opened 2020-03-24
Added snapshot file download when no local snapshot is found

alexsporn:


plugins/snapshot/download.go
```diff
@@ -0,0 +1,75 @@
+package snapshot
+
+import (
+	"fmt"
+	"github.com/dustin/go-humanize"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

muXxer (approved):

PR #365, opened 2020-04-01
Improves warp synchronization

luca-moser:
* Request missing approvees of milestones when doing milestone range requests
* Advance checkpoints everytime the solid milestone reaches over the previous checkpoint
* Reduce advancement range to 25 milestones
* Adds backpressure signal to request-background worker when the tangle processor is over half of its capacity. Note, if the tangle processor is reaching its capacity, the request queue will automatically have *a lot* of requests pending.
* Refactors warpsync.go package

During testing it was quite evident, that sync speed is now even more improved due to the checkpoint window shifting, but tests should only be concluded with the same static neighbors as speed is very much dependent on the peers:
* On mainnet a synchronization range of ~1000 milestones takes under a minute.
* On Comnet a synchronization range of ~1000 milestones would take around 8 minutes (note that Comnet runs at a steady ~80-150 TPS sometimes), but it depends heavily on the peers.

### DIFF
--- a/pkg/dag/helper.go
+++ b/pkg/dag/helper.go
@@ -57,3 +57,74 @@ func FindAllTails(txHash trinary.Hash, forceRelease bool) (map[string]struct{}, 
 	}
 	return tails, nil
 }
+
+// Predicate defines the condition whether the traversal should continue or not
+type Predicate func(cachedTx *tangle.CachedTransaction) bool
+
+// Consumer consumes the given transaction doing traversal
+type Consumer func(cachedTx *tangle.CachedTransaction)
+
+// OnMissingApprovee gets called when during traversal an approvee is missing.
+type OnMissingApprovee func(approveeHash trinary.Hash)
+
+// TraverseApprovees starts to traverse the approvees of the given start transaction until
+// the traversal stops due to transaction not passing the given condition.
+func TraverseApprovees(startTxHash trinary.Hash, condition Predicate, consumer Consumer, onMissingApprovee OnMissingApprovee) {
+
+	if tangle.SolidEntryPointsContain(startTxHash) {
+		return
+	}
+
+	processed := map[trinary.Hash]struct{}{}
+	txsToTraverse := map[trinary.Hash]struct{}{startTxHash: {}}
+	for len(txsToTraverse) != 0 {
+		for txHash := range txsToTraverse {
+			delete(txsToTraverse, txHash)
+
+			cachedTx := tangle.GetCachedTransactionOrNil(txHash) // tx +1
+			if cachedTx == nil {
+				continue
+			}
+
+			if txHash != startTxHash && !condition(cachedTx.Retain()) { // tx + 1
+				cachedTx.Release()
+				continue
+			}
+
+			// do not consume the start transaction
+			if txHash != startTxHash {
+				consumer(cachedTx.Retain()) // tx +1
+			}
+
+			approveeHashes := map[trinary.Hash]struct{}{
+				cachedTx.GetTransaction().GetTrunk():  {},
+				cachedTx.GetTransaction().GetBranch(): {},
+			}
+
+			cachedTx.Release(true) // tx -1
+
+			for approveeHash := range approveeHashes {
+				if tangle.SolidEntryPointsContain(approveeHash) {
+					continue
+				}
+
+				if _, checked := processed[approveeHash]; checked {
+					continue
+				}
+
+				processed[approveeHash] = struct{}{}
+
+				cachedApproveeTx := tangle.GetCachedTransactionOrNil(approveeHash) // approvee +1
+				if cachedApproveeTx == nil {
+					onMissingApprovee(approveeHash)
+					continue
+				}
+
+				// do not force release since it is loaded again
+				cachedApproveeTx.Release() // approvee -1
+
+				txsToTraverse[approveeHash] = struct{}{}
+			}
+		}
+	}
+}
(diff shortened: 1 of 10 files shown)


pkg/protocol/rqueue/rqueue.go
```diff
@@ -154,14 +155,21 @@ func (pq *priorityqueue) Received(hash trinary.Hash) *Request {
 			pq.latencySum = 0
 			pq.avgLatency.Store(0)
 		}
+
+		return req
 	}
-	pq.Unlock()
+
+	// check if the request is in the queue (was enqueued again after request)
+	req, _ := pq.queued[hash]
```
golangcibot: S1005: should write `req := pq.queued[hash]` instead of `req, _ := pq.queued[hash]` (from `gosimple`)

pkg/dag/helper.go
```diff
@@ -57,3 +57,74 @@ func FindAllTails(txHash trinary.Hash, forceRelease bool) (map[string]struct{},
...
+		for txHash := range txsToTraverse {
+			delete(txsToTraverse, txHash)
+
+			cachedTx := tangle.GetCachedTransactionOrNil(txHash) // tx +1
+			if cachedTx == nil {
+				continue
+			}
+
+			if txHash != startTxHash && !condition(cachedTx.Retain()) { // tx + 1
+				cachedTx.Release()
```
muXxer: Can you pass a parameter `forceRelease` ?

pkg/protocol/warpsync/warpsync.go
```diff
@@ -53,85 +49,133 @@ type Events struct {
...
+// DefaultAdvancementThreshold is the default threshold at which a checkpoint advancement is done.
+// Per default an advancement is always done as soon the solid milestone enters the range between
+// the previous and current checkpoint.
+const DefaultAdvancementThreshold = 0.0
+
+// AdvanceAtPercentReached is an AdvanceCheckpointCriteria which advances the checkpoint
+// when the current one was reached by >=X% by the current solid milestone in relation to the previous checkpoint.
+func AdvanceAtPercentReached(threshold float64) AdvanceCheckpointCriteria {
+	return func(currentSolid, previousCheckpoint, currentCheckpoint milestone.Index) bool {
+		// the previous checkpoint can be over the current solid one, as advancements
```
muXxer: Is there a word missing?

plugins/gossip/request.go
```diff
@@ -150,3 +158,37 @@ func RequestMilestoneApprovees(cachedMsBndl *tangle.CachedBundle) bool {
...
+			log.Panicf("milestone %d wasn't found", ms)
+		}
+
+		msBundleTailHash := cachedMsBundle.GetBundle().GetTailHash()
+		cachedMsBundle.Release(true) // bundle -1
+
+		dag.TraverseApprovees(msBundleTailHash,
+			// predicate
+			func(cachedTx *tangle.CachedTransaction) bool { // tx +1
+				defer cachedTx.Release() // tx -1
```
muXxer: Shouldn't we force release here?

plugins/gossip/request.go
```diff
@@ -150,3 +158,37 @@ func RequestMilestoneApprovees(cachedMsBndl *tangle.CachedBundle) bool {
...
+				_, previouslyTraversed := traversed[cachedTx.GetTransaction().GetHash()]
+				return !cachedTx.GetMetadata().IsSolid() && !previouslyTraversed
+			},
+			// consumer
+			func(cachedTx *tangle.CachedTransaction) { // tx +1
+				defer cachedTx.Release() // tx -1
```
muXxer: Shouldn't we force release here?

plugins/tangle/tangle_processor.go
```diff
@@ -112,20 +116,20 @@ func processIncomingTx(incomingTx *hornet.Transaction, request *rqueue.Request,
 	// Release shouldn't be forced, to cache the latest transactions
 	defer cachedTx.Release(!isNodeSyncedWithThreshold) // tx -1
 
+	// since we only add the approvees if there was a source request, we only
+	// request them for transactions which should be part of milestone cones
+	if request != nil {
```
muXxer: Move that function back into the `!alreadyAdded` condition, since this works now with the fix in rqueue for me.
(but I only tested with my `processing` map branch)

plugins/warpsync/plugin.go
```diff
@@ -20,7 +20,7 @@ import (
 var (
 	PLUGIN   = node.NewPlugin("WarpSync", node.Enabled, configure)
 	log      *logger.Logger
-	warpSync = warpsync.New(300)
+	warpSync = warpsync.New(25)
```
muXxer, line 23: Maybe we should add a setting in the config? What do you think?

muXxer (changes requested):

plugins/warpsync/plugin.go
luca-moser: lets leave it like this for now, otherwise people are going to mess with it too much

muXxer: :+1:

plugins/tangle/tangle_processor.go
luca-moser: moved

muXxer (approved):

plugins/gossip/request.go
luca-moser: ok

plugins/gossip/request.go
luca-moser: ok

pkg/protocol/warpsync/warpsync.go
luca-moser: added 'milestone'

pkg/dag/helper.go
luca-moser: added

PR #369, opened 2020-04-02
Fixes re-validations not working in conjunction with warp sync

luca-moser:
Makes the re-validation delete all the data above the last local snapshot to ensure a clean state of the database.

Fixes a bug where the heartbeat message wasn't updated on the peer on startup but it would trigger a warp sync in the same goroutine, causing milestone requests for the given start range to not actually being requested on the peer.


muXxer (approved):

plugins/tangle/solidifier.go
```diff
@@ -113,12 +110,10 @@ func checkSolidity(cachedTx *tangle.CachedTransaction) (solid bool, newlySolid b
...
 // Missing tx are requested
 // Can be aborted with abortSignal
-func solidQueueCheck(milestoneIndex milestone.Index, cachedMsTailTx *tangle.CachedTransaction, revalidate bool, abortSignal chan struct{}) (solid bool, aborted bool) {
+func solidQueueCheck(milestoneIndex milestone.Index, cachedMsTailTx *tangle.CachedTransaction, abortSignal chan struct{}) (solid bool, aborted bool) {
 
 	ts := time.Now()
-
-	solidMilestoneIndex := tangle.GetSolidMilestoneIndex()
-
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

PR #399, opened 2020-04-21
Fix wereAddressesSpentFrom with empty addresses

viossat:
# Description

Requesting `wereAddressesSpentFrom` with an empty `addresses` array leads to three issue:
- return a "No addresses provided" error, but an empty array should be valid (!= missing `addresses` parameter)
- return both the error and a result because a `return` statement is missing
- return `states` equals to `null` instead of `[]`

This PR fixes these issues.

### Actual result
```json
{
  "error": "No addresses provided"
}
{
  "states": null,
  "duration": 0
}
```

### Expected result
```json
{
  "states": [],
  "duration": 0
}
```

## Type of change

- [ ] Enhancement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

# How Has This Been Tested?

Basic requests with or without addresses.

# Checklist:

### DIFF
--- a/plugins/webapi/spent.go
+++ b/plugins/webapi/spent.go
@@ -39,12 +39,13 @@ func wereAddressesSpentFrom(i interface{}, c *gin.Context, _ <-chan struct{}) {
 		return
 	}
 
-	if len(query.Addresses) == 0 {
+	if query.Addresses == nil {
 		e.Error = "No addresses provided"
 		c.JSON(http.StatusBadRequest, e)
+		return
 	}
 
-	result := WereAddressesSpentFromReturn{}
+	result := WereAddressesSpentFromReturn{States: []bool{}}
 
 	for _, addr := range query.Addresses {
 		if err := address.ValidAddress(addr); err != nil {


plugins/webapi/spent.go
```diff
@@ -39,12 +39,13 @@ func wereAddressesSpentFrom(i interface{}, c *gin.Context, _ <-chan struct{}) {
 		return
 	}
 
-	if len(query.Addresses) == 0 {
+	if query.Addresses == nil {
```
IoTMOD, line 42: Please revert this change, because an empty query is a bad request

plugins/webapi/spent.go
```diff
@@ -39,12 +39,13 @@ func wereAddressesSpentFrom(i interface{}, c *gin.Context, _ <-chan struct{}) {
...
 		e.Error = "No addresses provided"
 		c.JSON(http.StatusBadRequest, e)
+		return
 	}
 
-	result := WereAddressesSpentFromReturn{}
+	result := WereAddressesSpentFromReturn{States: []bool{}}
```
IoTMOD, line 48: Please revert this, too

IoTMOD (changes requested):
Thanks for your PR!
The missing `return` is indeed an issue and should get fixed :ok_hand:

muXxer:
Sorry Mathieu... not today :)
But the missing return got fixed in another PR, thanks for that! :+1:

PR #409, opened 2020-04-29
add hornet helper scripts

bernardoaraujor:
add the following helper scripts:
 - `hornet_dashboard`
 - `hornet_clean_db`
 - `hornet_clean_export`

they're installed in `/usr/bin` and can be used as regular shell commands.


bernardoaraujor:
done in https://github.com/honeycombOS/hornetctl

PR #413, opened 2020-04-29
Fix possible overflow in balance calculation

muXxer:


pkg/model/tangle/bundle.go
```diff
@@ -321,6 +323,12 @@ func (bundle *Bundle) calcLedgerChanges() {
 		cachedTx := loadBundleTxIfExistsOrPanic(txHash, bundle.hash) // tx +1
 		if value := cachedTx.GetTransaction().Tx.Value; value != 0 {
 			changes[cachedTx.GetTransaction().Tx.Address] += value
+			if math.Abs(changes[cachedTx.GetTransaction().Tx.Address]) > int64(compressed.TotalSupply) {
```
luca-moser: This check is wrong, it shouldn't check whether one address goes over/below the supply but whether the bundle itself is spending/depositing more than the total supply. For example, just with this check, you could have a bundle with multiple input addresses which themselves are less than the total supply but their aggregation is still above the total supply: i.e. input A spends `-2779530283277760` and input B spends `-2779530283277760` which results in this check not doing anything.

So instead of doing this check, accumulate the bundle value (from input and outputs) and on each added/subtracted value, check whether the bundle value goes below or above the total supply.

Check IRI's bundle validation code which does exactly what I described above: https://github.com/iotaledger/iri/blob/adb57b97fbe52b5ca1095ca37b221e14ce0896bb/src/main/java/com/iota/iri/BundleValidator.java#L318

Also, this logic should be in the function validating the bundle not in `calcLedgerChanges()`

plugins/tangle/consistency.go
```diff
@@ -166,6 +177,10 @@ func computeConeDiff(visited map[trinary.Hash]struct{}, tailTxHash trinary.Hash,
 				ledgerChanges := cachedBndl.GetBundle().GetLedgerChanges()
 				for addr, change := range ledgerChanges {
 					coneDiff[addr] += change
+					if math.Abs(coneDiff[addr]) > int64(compressed.TotalSupply) {
+						// referenced bundle is not valid because ledger changes would overflow total supply
+						return nil, ErrRefBundleNotValid
```
luca-moser: Make a separate error because it isn't true that the bundle is invalid, it is the cone diff which isn't consistent.

plugins/tangle/consistency.go
```diff
@@ -63,6 +69,11 @@ func CheckConsistencyOfConeAndMutateDiff(tailTxHash trinary.Hash, approved map[t
 		// apply the latest ledger state's balance of the given address to the cone diff
 		change += int64(currentLedgerBalance)
 
+		if math.Abs(change) > int64(compressed.TotalSupply) {
+			// transaction is not consistent with the current diff because ledger changes would overflow total supply
```
luca-moser:
```suggestion
			// the mutation is not consistent with the current diff because the address would overflow/underflow from the total supply
```

plugins/tangle/consistency.go
```diff
@@ -49,6 +51,10 @@ func CheckConsistencyOfConeAndMutateDiff(tailTxHash trinary.Hash, approved map[t
 	// apply the walker diff to the cone diff
 	for addr, change := range diff {
 		coneDiff[addr] += change
+		if math.Abs(coneDiff[addr]) > int64(compressed.TotalSupply) {
+			// transaction is not consistent with the current diff because ledger changes would overflow total supply
```
luca-moser:
```suggestion
```

luca-moser (changes requested):

pkg/model/tangle/bundle.go
muXxer: Should I add this check in iota.go `ValidBundle()`? We already calculate the totalSum of the bundle there.

plugins/tangle/consistency.go
muXxer: We still need to add the bundle in the `refsAnInvalidBundleStorage` right?
So it would look like that?
```diff
        // compute the diff of the cone which the transaction references
	coneDiff, err := computeConeDiff(visited, tailTxHash, tangle.GetSolidMilestoneIndex(), forceRelease)
	if err != nil {
-		if err == ErrRefBundleNotValid {
+		if err == ErrRefBundleNotValid || err == ErrConeDiffNotConsistent {
			// memorize for a certain time that this transaction references an invalid bundle
			// to short circuit validation during a subsequent tip-sel on it again
			PutInvalidBundleReference(tailTxHash)
		}
		return false
	}
```

plugins/tangle/consistency.go
muXxer: Should I just remove the comment or use the same like in the other suggestion?
`// the mutation is not consistent with the current diff because the address would overflow/underflow from the total supply`

plugins/tangle/consistency.go
luca-moser: Yeah something like this 👍

plugins/tangle/consistency.go
luca-moser: Just remove the comment.

pkg/model/tangle/bundle.go
luca-moser: > Should I add this check in iota.go ValidBundle()?

Yes. 

Just to clarify, the changes in `ValidBundle()` should both include this check and include the validations within each iteration for `totalSum` like IRI does it.

muXxer: Check was added in iota.go

luca-moser (approved):

PR #414, opened 2020-04-30
feat: Update README

IoTMOD:


muXxer (approved):

ericvrp (commented):
Is RPi 4 the minimum or recommended minimum requirement? We can be a bit clearer about that.

IoTMOD:
The RPi 4 (4GB) is the recommended minimum requirement. 
It may work on a RPi 3B, but since the RPi 3B only has 1GB RAM, you may end in OOMs (especially if you try to run a Comnet node)

PR #417, opened 2020-05-03
Fix reuse of slice header in unsafe operation

muXxer:
This fixes the problem that nodes could not synchronize after a database re-validation.

### DIFF
--- a/plugins/tangle/revalidation.go
+++ b/plugins/tangle/revalidation.go
@@ -123,6 +123,8 @@ func cleanupTransactions(info *tangle.SnapshotInfo) {
 	start := time.Now()
 	var txCounter int64
 	tangle.ForEachTransactionHashBytes(func(txHashBytes []byte) {
+		txHashBytesCopy := make([]byte, 49)
+		copy(txHashBytesCopy, txHashBytes)
 
 		txCounter++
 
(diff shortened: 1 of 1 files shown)


IoTMOD (approved):

plugins/tangle/revalidation.go
```diff
@@ -123,29 +123,31 @@ func cleanupTransactions(info *tangle.SnapshotInfo) {
...
 			log.Infof("analyzed %d transactions", txCounter)
 		}
 
-		storedTxMeta := tangle.GetStoredMetadataOrNil(txHashBytes)
+		storedTxMeta := tangle.GetStoredMetadataOrNil(txHashBytesCopy)
 
 		// delete transaction if no metadata
 		if storedTxMeta == nil {
-			txsToDelete[typeutils.BytesToString(txHashBytes)] = struct{}{}
+			txsToDelete[typeutils.BytesToString(txHashBytesCopy)] = struct{}{}
```
Wollac, line 139: Just a remark: There is no need to call `BytesToString` for map keys. The go compiler assures that 
for `x = m[string(k)]` where `k` is `[]byte` no allocation or copying overhead occurs.
See https://github.com/golang/go/commit/f5f5a8b6209f84961687d993b93ea0d397f5d5bf for example

PR #429, opened 2020-05-11
Feature: Adding "sn_trytes" in zmq

MichiKuhl:
Publishes confirmed transactions trytes to zmq as already requested in the following IRI feature request: 
https://github.com/iotaledger/iri/pull/1551

### DIFF
--- a/config.json
+++ b/config.json
@@ -106,7 +106,7 @@
     "alias": "",
     "showAliasInGetNodeInfo": false,
     "disablePlugins": [],
-    "enablePlugins": ["zmq"]
+    "enablePlugins": []
   },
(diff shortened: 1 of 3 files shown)


IoTMOD:
Hello @MichiKuhl,

thank you for your PR. 

I don't know why this topic was added in IRI, because it introduces a similar issue that was recently closed in IRI and HORNET:
https://github.com/gohornet/hornet/pull/175 (if you subscribe to `sn`, you automatically subscribe to `sn_trytes` too, because it's the same prefix).

So if we really need the topic (I don't see a use case for it, to be honest), we should think about an other topic name.

PR #431, opened 2020-05-12
Spammer maxCPUUsage support

ericvrp:
Spamming will be greatly reducted when a cpu is used for other things.

This allows for less adjusting of the spammer in config.json . When the spammer is enabled it uses 50% of total cpu power on average by default. I run it at 90% on a 4 core cpu (8 threads, max 7 spammer workers) while keeping the system very responsive.

I would appreciate a critical review of the code since it is my first serious Golang code. (having experience in many other languages)


ericvrp:
P.S. Ideally I would have have cpu_usage.go moved to hive.go's runtime module

config.json
```diff
@@ -121,6 +121,7 @@
     "depth": 1,
     "message": "Spamming with HORNET tipselect",
     "tag": "HORNET99INTEGRATED99SPAMMER",
+    "maxCPUUsage": 0.5,
```
muXxer: Maybe `maxCpuUsage` looks better? I don't know :)

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+	"time"
+)
+
+var (
+	CPUUsageTimePerSample = time.Second / 4
+
+	activeWorkerCount = 0
+
+	once sync.Once
+	mu   sync.Mutex
```
muXxer: please use `syncutils.Mutex`  and give the mutex a name like `cpuUsageErrorLock`

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+
+	cpu_last_sum uint64   // previous iteration
```
muXxer: Underlines are not used in golang for variable names

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+
+	cpuUsageResult float64 // current result
+	cpuUsageError  error   // nil || current error
+)
+
+func cpuUsageUpdater() {
+	if runtime.GOOS == "windows" {
+		mu.Lock()
+		defer mu.Unlock()
+		cpuUsageError = errors.New("spammer.maxCPUUsage on Windows is not supported")
```
muXxer: You should panic with this error in the `configure` func of the spammer plugin. No need for this lock, no need for this global error variable.

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+
+	go func() {
+		for {
+			procStat, err := ioutil.ReadFile("/proc/stat")
```
muXxer: Don't use that reading of `/proc/stat`.
We already use the `psutil` package somewhere else. Just use the `Percent()` function in that package.

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+					mu.Lock()
+					defer mu.Unlock()
+					cpuUsageError = errors.New("Can't convert from string to int")
+					return
+				}
+				cpu_sum += n
+				cpu_now[i] = n
+			}
+
+			if len(cpu_last) != 0 { // not on first iteration
```
muXxer: psutil package also contains the history of last stats already

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+	mu   sync.Mutex
+
+	cpu_last_sum uint64   // previous iteration
+	cpu_last     []uint64 // previous iteration
+
+	cpuUsageResult float64 // current result
+	cpuUsageError  error   // nil || current error
+)
+
+func cpuUsageUpdater() {
```
muXxer, line 26: Please comment that function like:
``` 
// FunctionName does blabla.
func FunctionName() { 
```

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+
+			cpu_last_sum = cpu_sum
+			cpu_last = cpu_now
+
+			time.Sleep(CPUUsageTimePerSample)
+		} // next for...
+	}()
+}
+
+func CPUUsageAdjust(nWorkerChange int) { // preemptive adjust cpu usage so we don't idle/run all workers at once
```
muXxer: Please comment that function like:
``` 
// FunctionName does blabla.
func FunctionName() { 
```

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+	activeWorkerCount += nWorkerChange
+	// cpuUsageResultOld := cpuUsageResult
+	cpuUsageResult += float64(nWorkerChange) / float64(runtime.NumCPU()) // assume one core usage per worker
+	// log.Infof("CPUUsageAdjust active workers with %d to %d. cpu %.2f -> %.2f",
+	// 	nWorkerChange, activeWorkerCount,
+	// 	cpuUsageResultOld, cpuUsageResult)
+	mu.Unlock()
+}
+
+func CPUUsage() (float64, error) { // see: https://www.idnt.net/en-GB/kb/941772
```
muXxer: Please comment that function like:
``` 
// FunctionName does blabla.
func FunctionName() { 
```

plugins/spammer/cpu_usage.go
```diff
@@ -0,0 +1,107 @@
...
+	once.Do(cpuUsageUpdater)
+
+	mu.Lock()
+	defer mu.Unlock()
+	return cpuUsageResult, cpuUsageError // XXX mu.Lock() around this?
+}
+
+func randomSleep() { // prefend gettings into a cpu eating loop
```
muXxer: Please comment that function like:
``` 
// FunctionName does blabla.
func FunctionName() { 
```

plugins/spammer/plugin.go
```diff
@@ -48,8 +50,20 @@ func configure(plugin *node.Plugin) {
...
+		} else {
+			rateLimit = 0.0
+			spammerWorkerCount = runtime.NumCPU() - 1 // note: having more workers in this case actually seems to make the cpu usage more stable. Keeping the amount 'sane' for now though
+			// log.Infof("maxCPUUsage=%.2f rateLimit=%.2f spammerWorkerCount=%d", maxCPUUsage, rateLimit, spammerWorkerCount)
+		}
+	}
+
 	if rateLimit != 0 {
-		rateLimitChannelSize := int64(rateLimit) * 2
+		rateLimitChannelSize := int64(rateLimit) * 2 // XXX this case looks wrong and I doubt this code works as intended
```
muXxer: Why? It creates a channel with double the size of the rate limit. No bigger size needed.

plugins/spammer/spammer.go
```diff
@@ -53,11 +87,15 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
 	b, err := createBundle(address, message, tagSubstring, txCountValue, infoMsg)
 	if err != nil {
+		log.Infof("worker idle because createBundle err: %s", err)
+		randomSleep()
 		return
 	}
 
 	err = doPow(b, tips[0], tips[1], mwm)
 	if err != nil {
+		log.Infof("worker idle because doPow err: %s", err)
```
muXxer: Remove these comments.

plugins/spammer/spammer.go
```diff
@@ -53,11 +87,15 @@ func doSpam(shutdownSignal <-chan struct{}) {
 
 	b, err := createBundle(address, message, tagSubstring, txCountValue, infoMsg)
 	if err != nil {
+		log.Infof("worker idle because createBundle err: %s", err)
```
muXxer: Remove these comments.

plugins/spammer/spammer.go
```diff
@@ -67,13 +105,15 @@ func doSpam(shutdownSignal <-chan struct{}) {
 	for _, tx := range b {
 		txTrits, _ := transaction.TransactionToTrits(&tx)
 		if err := gossip.Processor().CompressAndEmit(&tx, txTrits); err != nil {
+			log.Infof("worker idle because CompressAndEmit err: %s", err)
```
muXxer: Remove these comments.

plugins/spammer/spammer.go
```diff
@@ -67,13 +105,15 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
+			randomSleep()
 			return
 		}
 		metrics.SharedServerMetrics.SentSpamTransactions.Inc()
 	}
 
 	durTotal := time.Since(timeStart).Truncate(time.Millisecond)
-	log.Infof("Sent Spam Transaction: #%d, TxHash: %v, GTTA: %v, PoW: %v, Total: %v", txCountValue, b[0].Hash, durGTTA.Truncate(time.Millisecond), durPOW.Truncate(time.Millisecond), durTotal.Truncate(time.Millisecond))
+	log.Infof("sent transaction: #%d, TxHash: %v, GTTA: %v, PoW: %v, Total: %v", txCountValue, b[0].Hash, durGTTA.Truncate(time.Millisecond), durPOW.Truncate(time.Millisecond), durTotal.Truncate(time.Millisecond))
```
muXxer: `sent spam transaction` is the correct comment.
Only `sent transaction` could be confusing IMHO.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
 	}
 
 	if !tangle.IsNodeSyncedWithThreshold() {
+		// log.Infof("worker idle because: !tangle.IsNodeSyncedWithThreshold()")
+		randomSleep()
 		return
 	}
 
+	if !tangle.IsNodeSynced() {
```
muXxer: This check is definitely wrong. Remove it.

plugins/spammer/cpu_usage.go
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
 	}
 
 	if !tangle.IsNodeSyncedWithThreshold() {
+		// log.Infof("worker idle because: !tangle.IsNodeSyncedWithThreshold()")
+		randomSleep()
```
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
 		return
 	}
 
+	if !tangle.IsNodeSynced() {
+		// log.Infof("worker idle because: !tangle.IsNodeSynced()")
+		randomSleep()
```
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
+		return
+	}
+
+	if peering.Manager().ConnectedPeerCount() == 0 {
+		// log.Infof("worker idle because: peering.Manager().ConnectedPeerCount() == 0")
+		randomSleep()
```
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
+		return
+	}
+
+	if maxCPUUsage > 0.0 {
+		cpuUsage, err := CPUUsage()
+		if err == nil {
+			// log.Infof("cpuUsage %.2f\n", cpuUsage)
+			if cpuUsage > maxCPUUsage {
+				// log.Infof("worker idle with cpuUsage %.2f > %.2f", cpuUsage, maxCPUUsage)
+				randomSleep()
```
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
+	}
+
+	CPUUsageAdjust(+1)
+	defer CPUUsageAdjust(-1)
+
 	timeStart := time.Now()
 	tips, _, err := tipselection.SelectTips(depth, nil)
 	if err != nil {
+		log.Infof("worker idle because SelectTips err: %s", err)
+		randomSleep()
```
muXxer: Why do we even need these sleeps at all? Just abort the `doSpam` without the sleeps, because it is triggered by a ticker anyway.

plugins/spammer/spammer.go
```diff
@@ -37,12 +38,45 @@ func doSpam(shutdownSignal <-chan struct{}) {
...
+		}
+	}
+
+	CPUUsageAdjust(+1)
+	defer CPUUsageAdjust(-1)
+
 	timeStart := time.Now()
 	tips, _, err := tipselection.SelectTips(depth, nil)
 	if err != nil {
+		log.Infof("worker idle because SelectTips err: %s", err)
```
muXxer: Remove these comments.

plugins/spammer/plugin.go
```diff
@@ -48,8 +50,20 @@ func configure(plugin *node.Plugin) {
 		spammerWorkerCount = 1
 	}
 
+	if maxCPUUsage > 0.0 {
+		_, err := CPUUsage()
```
muXxer: Do something like:
`if _, err := CPUUsage(); err != nil {`

muXxer (changes requested):

muXxer:
> P.S. Ideally I would have have cpu_usage.go moved to hive.go's runtime module

If it is not needed somewhere else, lets keep it in the spammer.

config.json
ericvrp: Yes, the CPUU in the middle is hard to read. What I did is the rules/guidlines from "The Go Programming Language" book that say letter if acronyms and initialisms like ASCII and HTML are always rendered in the same case. So perhaps cpuMaxUsage ?

muXxer: cpuMaxUsage sounds good

plugins/spammer/cpu_usage.go
ericvrp: Alright. Yes I need to get to know more what is in hive.go . Same goes for sync.Once -> syncutils.Once I assume.

P.S. Why is there a syncutils when in this case it just says?
> type Mutex = sync.Mutex

muXxer: We have that as a placeholder, if we want to enable the "deadlock" detector, we just replace that line with deadlock.Mutex. This way we don't need to change all mutexes in the whole codebase.

plugins/spammer/plugin.go
ericvrp: My comment should have been removed before checking in.

Do we need the rateLimit channel to be buffered?

plugins/spammer/spammer.go
ericvrp: Now the log output looks like this
`0|hornet  | 2020-05-13T12:35:14+02:00   INFO    Spammer sent transaction: #958153, TxHash: KAZESYBZCY9D9CNBQJKOQDZRRKUUJM9VPGDLFNWMFZSLOKIOGSUWSYJOGUTWZWHJTTL9XFCCMYZWV9999, GTTA: 111ms, PoW: 186ms, Total: 298ms`

In the "Sent Spam Transaction" I had the feeling the Spam word was double in the final output. Plus the line was longer and wrapped more easy

plugins/spammer/cpu_usage.go
ericvrp: The goal of cpuMaxUsage is to not have to set a rateLimit. Because we now use the CPUUsageAdjust(...) functionality we might have no noticable cpu usage when we set rateLimit to something high like 250tps. But I would prefer not to guesstimate these things.

plugins/spammer/cpu_usage.go
ericvrp: That is very cool. I'm reading the deadlock package docs...

plugins/spammer/spammer.go
muXxer: fair enough! :+1:

IoTMOD (approved):

muXxer (approved):
Hey Eric!
Thanks for your pull request. I changed some things in your code. Maybe you can have a look at the final result.

ericvrp:
I like that there are fewer changes in the core of the code (doSpam).

The current code doesn't work as good as the previous version though. I think the new code that guesses the new cpu usage by assuming one more used core needs to keep into account when spammers start and stop. Per start we used to assume one core used and vise versa. 

When in the current code we for instance start with 0.1 cpu usage and all X workers start at once assuming there will be 0.1+1/nCpus . In my code the first worker would assume 0.1+X/nCpus, the second 0.1+(X+1)/nCpus, etc. with X being nStarted-nStopped workers since last update of cpu_usage (0.1 in this example). This gives a much better estimation of the actual cpu_usage and would keep the cpu usage at the requested value. In the current version the cpu usage fluctuates between 0.43 and 0.64 and doesn't get near the requested max of 0.8

PR #442, opened 2020-05-20
Speed up revalidation and pruning.

muXxer:
We do not access the storage layer directly anymore.
This way we can use the batched writes of the object storage to speed up storage layer access.


pkg/model/tangle/unconfirmed_tx_storage.go
```diff
@@ -97,19 +97,16 @@ func StoreUnconfirmedTx(msIndex milestone.Index, txHash trinary.Hash) *CachedUnc
...
 
 	msIndexBytes := make([]byte, 4)
 	binary.LittleEndian.PutUint32(msIndexBytes, uint32(msIndex))
 
-	var txHashes [][]byte
 	unconfirmedTxStorage.ForEachKeyOnly(func(key []byte) bool {
-		txHashes = append(txHashes, key)
+		unconfirmedTxStorage.Delete(key)
 		return true
 	}, true, msIndexBytes)
```
alexsporn: change skipCache to false?

muXxer: Fair enough

alexsporn (approved):

PR #479, opened 2020-06-04
feat: Dashboard improvements

### LINKED ISSUE #472: Dashboard improvements
When looking at the dashboard I find myself doing three things over and over again. 

1. When my node restarts Hornet says my node is synced while it has not received any milestone update from neighbors yet. I understand it thinks it has caught up but it could be a bit smarter.
  proposed improvement: give this state a different color and text instead of saying synced in green.

2. Figuring out how many milestones I am behind.
  proposed improvement: say something like "1234 to LMI 111791 " (same on neighbors page?) 

3. When it says "Connected Neighbors: 3" I am constantly switching to the Neighbor page to see what their status is.
  proposed improvement: add a colored square (after the number 3) indicating the status of each peer. Preferably with inbound/outbound indicators.

IoTMOD:
closes #472

- Improves the sync status indicator. A node is considered to be sync if `lmi == lsmi`, it has active peers (at least one) and the lmi is not too old
- Adds a tooltip to show the milestone delta if the node is not synced
image

### DIFF
--- a/pkg/utils/health.go
+++ b/pkg/utils/health.go
@@ -1,0 +1,40 @@
+package utils
+
+import (
+	"time"
+
+	"github.com/gohornet/hornet/pkg/model/tangle"
+	"github.com/gohornet/hornet/plugins/peering"
+)
+
+const maxAllowedMilestoneAge = time.Minute * 5
+
+// IsNodeHealthy returns whether the node is synced, has active neighbors and its latest milestone is not too old
+func IsNodeHealthy() bool {
+	// Synced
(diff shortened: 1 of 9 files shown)


pkg/utils/health.go
```diff
@@ -0,0 +1,40 @@
+package utils
+
+import (
```
muXxer: Move this file to the tangle plugin

muXxer (changes requested):

muXxer (approved):

PR #510, opened 2020-06-10
Adds the ability to let the Prometheus plugin create a 'file service discovery' file

luca-moser:
The config opts are not exposed in config.json as they are rather exotic and not meant for the normal use case.

### DIFF
--- a/pkg/config/prometheus_config.go
+++ b/pkg/config/prometheus_config.go
@@ -13,7 +13,7 @@ const (
 	CfgPrometheusProcessMetrics = "prometheus.processMetrics"
 	// include promhttp metrics
 	CfgPrometheusPromhttpMetrics = "prometheus.promhttpMetrics"
-	// "whether the plugin should write a Prometheus 'file SD' file"
+	// whether the plugin should write a Prometheus 'file SD' file
 	CfgPrometheusFileServiceDiscoveryEnabled = "prometheus.fileServiceDiscovery.enabled"
 	// the path where to write the 'file SD' file to
 	CfgPrometheusFileServiceDiscoveryPath = "prometheus.fileServiceDiscovery.path"
@@ -27,6 +27,6 @@ func init() {
 	flag.Bool(CfgPrometheusProcessMetrics, false, "include process metrics")
 	flag.Bool(CfgPrometheusPromhttpMetrics, false, "include promhttp metrics")
 	flag.Bool(CfgPrometheusFileServiceDiscoveryEnabled, false, "whether the plugin should write a Prometheus 'file SD' file")
-	flag.String(CfgPrometheusFileServiceDiscoveryPath, "./target.json", "the path where to write the 'file SD' file to")
-	flag.String(CfgPrometheusFileServiceDiscoveryTarget, "127.0.0.1:9311", "the target to write into the 'file SD' file")
+	flag.String(CfgPrometheusFileServiceDiscoveryPath, "target.json", "the path where to write the 'file SD' file to")
+	flag.String(CfgPrometheusFileServiceDiscoveryTarget, "localhost:9311", "the target to write into the 'file SD' file")
 }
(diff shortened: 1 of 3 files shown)


pkg/config/prometheus_config.go
```diff
@@ -13,11 +13,20 @@ const (
 	CfgPrometheusProcessMetrics = "prometheus.processMetrics"
 	// include promhttp metrics
 	CfgPrometheusPromhttpMetrics = "prometheus.promhttpMetrics"
+	// "whether the plugin should write a Prometheus 'file SD' file"
```
muXxer: remove the ""

pkg/config/prometheus_config.go
```diff
@@ -13,11 +13,20 @@ const (
...
 )
 
 func init() {
 	flag.String(CfgPrometheusBindAddress, "localhost:9311", "the bind address on which the Prometheus exporter listens on")
 	flag.Bool(CfgPrometheusGoMetrics, false, "include go metrics")
 	flag.Bool(CfgPrometheusProcessMetrics, false, "include process metrics")
 	flag.Bool(CfgPrometheusPromhttpMetrics, false, "include promhttp metrics")
+	flag.Bool(CfgPrometheusFileServiceDiscoveryEnabled, false, "whether the plugin should write a Prometheus 'file SD' file")
+	flag.String(CfgPrometheusFileServiceDiscoveryPath, "./target.json", "the path where to write the 'file SD' file to")
+	flag.String(CfgPrometheusFileServiceDiscoveryTarget, "127.0.0.1:9311", "the target to write into the 'file SD' file")
```
muXxer: Does localhost also work?
We use localhost in the rest of the config.

pkg/config/prometheus_config.go
```diff
@@ -13,11 +13,20 @@ const (
...
+	CfgPrometheusFileServiceDiscoveryTarget = "prometheus.fileServiceDiscovery.target"
 )
 
 func init() {
 	flag.String(CfgPrometheusBindAddress, "localhost:9311", "the bind address on which the Prometheus exporter listens on")
 	flag.Bool(CfgPrometheusGoMetrics, false, "include go metrics")
 	flag.Bool(CfgPrometheusProcessMetrics, false, "include process metrics")
 	flag.Bool(CfgPrometheusPromhttpMetrics, false, "include promhttp metrics")
+	flag.Bool(CfgPrometheusFileServiceDiscoveryEnabled, false, "whether the plugin should write a Prometheus 'file SD' file")
+	flag.String(CfgPrometheusFileServiceDiscoveryPath, "./target.json", "the path where to write the 'file SD' file to")
```
muXxer: Would `target.json` without the `./` also work?

pkg/config/prometheus_config.go
```diff
@@ -13,11 +13,20 @@ const (
 	CfgPrometheusProcessMetrics = "prometheus.processMetrics"
 	// include promhttp metrics
 	CfgPrometheusPromhttpMetrics = "prometheus.promhttpMetrics"
+	// "whether the plugin should write a Prometheus 'file SD' file"
+	CfgPrometheusFileServiceDiscoveryEnabled = "prometheus.fileServiceDiscovery.enabled"
+	// the path where to write the 'file SD' file to
+	CfgPrometheusFileServiceDiscoveryPath = "prometheus.fileServiceDiscovery.path"
+	// the target to write into the 'file SD' file
+	CfgPrometheusFileServiceDiscoveryTarget = "prometheus.fileServiceDiscovery.target"
```
muXxer, line 21: Would it make sense to create a slice with several targets?

muXxer (changes requested):

pkg/config/prometheus_config.go
luca-moser: no

muXxer (approved):

PR #526, opened 2020-06-16
Fix below max depth for SolidEntryPoints

muXxer:
This PR fixes the "below max depth" calculation for solid entry points.

- SEP index is now checked if we hit a SEP during "below-max-depth" check. If we are below lowestAllowedSnapshotIndex, we are below max depth. If not, we are not.
- Networks must be bootstrapped with global snapshots to mark Genesis tx as SEP with correct index.
- Genesis Tx is not set as SEP anymore by local snapshots.

This PR also introduces a flag to force the loading of a global snapshot even if a database already exists. Old Txs are kept in the database, only SEP's and the ledger state is replaces with the state in the snapshot (snapshot index must be higher than LSMI in DB).

### DIFF
--- a/pkg/model/hornet/solid_entry_points.go
+++ b/pkg/model/hornet/solid_entry_points.go
@@ -36,6 +36,11 @@ func (s *SolidEntryPoints) Contains(txHash Hash) bool {
 	return exists
 }
 
+func (s *SolidEntryPoints) Index(txHash Hash) (bool, milestone.Index) {
+	index, exists := s.entryPointsMap[string(txHash)]
+	return exists, index
+}
+
 func (s *SolidEntryPoints) Add(txHash Hash, milestoneIndex milestone.Index) {
 	if _, exists := s.entryPointsMap[string(txHash)]; !exists {
 		s.entryPointsMap[string(txHash)] = milestoneIndex
--- a/pkg/model/hornet/solid_entry_points.go
+++ b/pkg/model/hornet/solid_entry_points.go
@@ -36,9 +36,9 @@ func (s *SolidEntryPoints) Contains(txHash Hash) bool {
 	return exists
 }
 
-func (s *SolidEntryPoints) Index(txHash Hash) (bool, milestone.Index) {
+func (s *SolidEntryPoints) Index(txHash Hash) (milestone.Index, bool) {
 	index, exists := s.entryPointsMap[string(txHash)]
-	return exists, index
+	return index, exists
 }
 
 func (s *SolidEntryPoints) Add(txHash Hash, milestoneIndex milestone.Index) {
--- a/pkg/model/tangle/solid_entry_points.go
+++ b/pkg/model/tangle/solid_entry_points.go
@@ -66,6 +66,16 @@ func SolidEntryPointsContain(txHash hornet.Hash) bool {
 	return solidEntryPoints.Contains(txHash)
 }
 
+func SolidEntryPointsIndex(txHash hornet.Hash) (bool, milestone.Index) {
+	ReadLockSolidEntryPoints()
+	defer ReadUnlockSolidEntryPoints()
+
+	if solidEntryPoints == nil {
+		panic(ErrSolidEntryPointsNotInitialized)
+	}
+	return solidEntryPoints.Index(txHash)
+}
+
 // WriteLockSolidEntryPoints must be held while entering this function
 func SolidEntryPointsAdd(txHash hornet.Hash, milestoneIndex milestone.Index) {
 	if solidEntryPoints == nil {
--- a/pkg/model/tangle/solid_entry_points.go
+++ b/pkg/model/tangle/solid_entry_points.go
@@ -66,7 +66,7 @@ func SolidEntryPointsContain(txHash hornet.Hash) bool {
 	return solidEntryPoints.Contains(txHash)
 }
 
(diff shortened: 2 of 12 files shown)


pkg/model/hornet/solid_entry_points.go
```diff
@@ -36,6 +36,11 @@ func (s *SolidEntryPoints) Contains(txHash Hash) bool {
 	return exists
 }
 
+func (s *SolidEntryPoints) Index(txHash Hash) (bool, milestone.Index) {
+	index, exists := s.entryPointsMap[string(txHash)]
+	return exists, index
```
alexsporn: why swap? go convention seems to be value, bool

pkg/model/tangle/solid_entry_points.go
```diff
@@ -66,6 +66,16 @@ func SolidEntryPointsContain(txHash hornet.Hash) bool {
 	return solidEntryPoints.Contains(txHash)
 }
 
+func SolidEntryPointsIndex(txHash hornet.Hash) (bool, milestone.Index) {
```
alexsporn: swap bool and index in the return value

pkg/model/hornet/solid_entry_points.go
muXxer: We have somewhere the `GetConfirmed()` which is also `bool, index`

pkg/model/tangle/solid_entry_points.go
alexsporn: yes, but those are 2 values we return there, not a "what you requested does not exist" boolean flag

pkg/model/hornet/solid_entry_points.go
alexsporn: yes, but those are 2 values we return there, not a "what you requested does not exist" boolean flag

pkg/model/hornet/solid_entry_points.go
```diff
@@ -36,6 +36,11 @@ func (s *SolidEntryPoints) Contains(txHash Hash) bool {
 	return exists
 }
 
+func (s *SolidEntryPoints) Index(txHash Hash) (milestone.Index, bool) {
+	index, exists := s.entryPointsMap[string(txHash)]
+	return index, exists
```
alexsporn, line 41: really?

muXxer: grafik
yes!

alexsporn (approved):

pkg/model/hornet/solid_entry_points.go
GalRogozinski: liked that german IDE

Whow, go has surprising behavior though

plugins/snapshot/local_snapshot.go
```diff
@@ -201,7 +201,6 @@ func shouldTakeSnapshot(solidMilestoneIndex milestone.Index) bool {
 func getSolidEntryPoints(targetIndex milestone.Index, abortSignal <-chan struct{}) (map[string]milestone.Index, error) {
 
 	solidEntryPoints := make(map[string]milestone.Index)
-	solidEntryPoints[string(hornet.NullHashBytes)] = targetIndex
```
GalRogozinski, line 204: So if I understand correctly, transactions referencing genesis in a hornet node that starts from a local snapshot will never be solid?

I guess it makes sense

plugins/snapshot/pruning.go
```diff
@@ -85,7 +84,7 @@ func pruneTransactions(txsToCheckMap map[string]struct{}) int {
 
 	for txHashToDelete := range txsToDeleteMap {
 
-		if txHashToDelete == consts.NullHashTrytes {
+		if bytes.Equal(hornet.Hash(txHashToDelete), hornet.NullHashBytes) {
```
GalRogozinski, line 87: Excuse my poor go skills
does the previous one fail under some condition?
I just want to understand the motivation for the change

plugins/tangle/depth.go
```diff
@@ -60,12 +60,12 @@ var BelowDepthMemoizationCache = belowDepthMemoizationCache{memoizationCache: ma
 // within the range of allowed milestones. if not, it is checked whether any referenced (directly/indirectly) tx
 // is confirmed by a milestone below the allowed threshold until a limit is reached of analyzed txs, in which case
 // the given tail transaction is also deemed being below max depth.
-func IsBelowMaxDepth(cachedTailTx *tangle.CachedTransaction, lowerAllowedSnapshotIndex int, forceRelease bool) bool {
+func IsBelowMaxDepth(cachedTailTx *tangle.CachedTransaction, lowestAllowedSnapshotIndex int, forceRelease bool) bool {
```
GalRogozinski, line 63: good name change :-)
I was looking at tip-sel before and I wrote a note that said `lower` was confusing

plugins/snapshot/local_snapshot.go
muXxer: Correct. Either the local snapshot contains the SEP for some reason, or we need a global snapshot.

plugins/snapshot/pruning.go
muXxer: `txHashToDelete` is the hash encoded in bytes, but `NullHashTrytes` is encoded in trytes. :)

PR #532, opened 2020-06-18
Add Snapshot index and Pruning index to Prometheus

c-mnd:
# Description

Adds Snapshot index as `iota_info_snapshot_index`to the Prometheus plugin.
Adds Pruning index as `iota_info_pruning_index` to the Prometheus plugin.

## Type of change

# How Has This Been Tested?

Both metrics show up in the Prometheus expression browser correctly.

# Checklist:


plugins/prometheus/info.go
```diff
@@ -105,6 +117,20 @@ func collectInfo() {
 		cachedSolidMs.Release()
 	}
 
+	snapshotInfo := tangle.GetSnapshotInfo()
+
+	// Snapshot index
+	if snapshotInfo != nil {
+		si := tangle.GetSnapshotInfo().SnapshotIndex
```
muXxer: You can set both values in the same if condition. No need to check again.

And you could set it directly, for example `infoSnapshotIndex.Set(float64(tangle.GetSnapshotInfo().SnapshotIndex))`

muXxer (changes requested):

plugins/prometheus/info.go
c-mnd: The style should follow the rest of this plugin, that's why I did it seperately.

muXxer: There is no line in the code where we check the same condition twice. Please change it ;)

c-mnd:
But yeah, will do so!

plugins/prometheus/info.go
c-mnd: But yeah, will do so!

c-mnd: But I'd rather use previously defined `snapshotInfo` var
`infoSnapshotIndex.Set(float64(snapshotInfo.SnapshotIndex))`
than calling `GetSnapshotInfo` again by
`infoSnapshotIndex.Set(float64(tangle.GetSnapshotInfo().SnapshotIndex))`
? right?

muXxer: Ah, sure! :)
Thats better!

muXxer (approved):

plugins/prometheus/info.go
c-mnd: Think I'm done with requested changes.
Now is it me or you @muXxer to click "Resolve conversation"? Not sure.

PR #547, opened 2020-06-30
fix: Stuck autopeers

IoTMOD:


pkg/peering/peering.go
```diff
@@ -263,6 +263,19 @@ func (m *Manager) ForAllConnected(f PeerConsumerFunc) {
 	}
 }
 
+// ForAll executes the given function for each peer until
+// abort is returned from within the consumer function.
+func (m *Manager) ForAll(f PeerConsumerFunc) {
+	m.RLock()
+	defer m.RUnlock()
+	for _, p := range m.connected {
```
muXxer, line 271: Please check if looping over reconnect pool is also needed.

pkg/peering/peering.go
```diff
@@ -263,6 +263,19 @@ func (m *Manager) ForAllConnected(f PeerConsumerFunc) {
...
+		if f(p) {
```
muXxer: Please flip the logic here and in `ForAllConnected` to abort on `false` instead of `true`.
This would match all other iterations in HORNET

pkg/protocol/bqueue/bqueue.go
```diff
@@ -54,13 +54,13 @@ func (bc *queue) Run(shutdownSignal <-chan struct{}) {
 		case b := <-bc.c:
 			bc.manager.ForAllConnected(func(p *peer.Peer) (abort bool) {
 				if _, excluded := b.ExcludePeers[p.ID]; excluded {
-					return
+					return false
```
muXxer: Must be `true`

pkg/protocol/bqueue/bqueue.go
```diff
@@ -54,13 +54,13 @@ func (bc *queue) Run(shutdownSignal <-chan struct{}) {
...
 				}
 
 				// just send the transaction when the peer supports STING
 				if p.Protocol.Supports(sting.FeatureSet) {
 					helpers.SendTransaction(p, b.TxData)
-					return
+					return false
```
muXxer: Must be `true`

muXxer (changes requested):

muXxer (approved):

PR #555, opened 2020-07-02
Remove graph and monitor plugin

muXxer:
This PR removes the graph and monitor plugins, since they don't fit the whiteflag approach anymore.


alexsporn (approved):

alexsporn:
Cleanup the configs too

PR #556, opened 2020-07-02
Remove legacy protocol support

muXxer:
This PR removes the support for legacy protocol messages.
This will finally break the compatibility to IRI :tada: :heart_eyes:


pkg/protocol/protocol.go
```diff
@@ -33,7 +32,7 @@ var (
 	*/
 
 	// supports protocol messages/feature sets: 2+1
```
alexsporn: fix comment

alexsporn (approved):

PR #557, opened 2020-07-03
Fix hiding of all non essential config flags

muXxer:


alexsporn (dismissed):

alexsporn:
when I test this branch I see this output:
```
./hornet -h 
Usage of ./hornet:
  -c, --config string                 Filename of the config file without the file extension (default "config")
  -d, --config-dir string             Path to the directory containing the config file (default ".")
      --node.disablePlugins strings   a list of plugins that shall be disabled
      --node.enablePlugins strings    a list of plugins that shall be enabled
      --overwriteCooAddress           apply new coordinator address from config file to database
  -n, --peeringConfig string          Filename of the peering config file without the file extension (default "peering")
      --profilesConfig string         Filename of the profiles config file without the file extension (default "profiles")
  -p, --useProfile string             Sets the profile with which the node runs (default "auto")
  -v, --version                       Prints the HORNET version
pflag: help requested
```

alexsporn (approved):

PR #564, opened 2020-07-16
Added hornet.Hash utils

alexsporn:
Added new `HashFromAddressTrytes`, `HashFromHashTrytes` and `HashFromTagTrytes` to the hornet pkg to properly convert and trim trytes to bytes

### DIFF
--- a/pkg/model/hornet/hash.go
+++ b/pkg/model/hornet/hash.go
@@ -14,6 +14,18 @@ var (
 // Hash is the binary representation of a trinary Hash.
 type Hash []byte
 
+func HashFromAddressTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes[:81])[:49]
+}
+
+func HashFromHashTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes[:81])[:49]
+}
+
+func HashFromTagTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes)[:17]
+}
+
 // Trytes converts the binary Hash to its trinary representation.
 func (h Hash) Trytes() trinary.Trytes {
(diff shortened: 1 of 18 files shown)


pkg/model/hornet/hash.go
```diff
@@ -14,6 +14,18 @@ var (
...
+func HashFromAddressTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes[:81])[:49]
+}
+
+func HashFromHashTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes[:81])[:49]
+}
+
+func HashFromTagTrytes(trytes trinary.Trytes) Hash {
+	return trinary.MustTrytesToBytes(trytes)[:17]
```
muXxer: why not cut the input to 27?

plugins/webapi/transactions.go
```diff
@@ -109,7 +109,7 @@ func findTransactions(i interface{}, c *gin.Context, _ <-chan struct{}) {
 			addr = addr[:81]
 		}
 
-		txHashes = append(txHashes, tangle.GetTransactionHashesForAddress(hornet.Hash(trinary.MustTrytesToBytes(addr)[:49]), query.ValueOnly, true, maxResults-len(txHashes)).Trytes()...)
+		txHashes = append(txHashes, tangle.GetTransactionHashesForAddress(hornet.HashFromHashTrytes(addr), query.ValueOnly, true, maxResults-len(txHashes)).Trytes()...)
```
muXxer: HashFromAddressTrytes

muXxer (changes requested):

pkg/model/hornet/hash.go
alexsporn: because I forgot how many trytes a tag has and then forgot to look it up

muXxer (approved):

PR #567, opened 2020-07-22
Adaptive heaviest branch tipsel

muXxer:
This PR implements the adaptive heaviest branch tipselection for the coordinator.

### DIFF
--- a/config.json
+++ b/config.json
@@ -77,7 +77,15 @@
     "stateFilePath": "coordinator.state",
     "merkleTreeFilePath": "coordinator.tree",
     "intervalSeconds": 60,
-    "checkpointTransactions": 5
+    "checkpoints": {
+      "maxApproveesCount": 10000,
+      "maxTipsCount": 100
+    },
+    "tipsel": {
+      "minHeaviestBranchUnconfirmedTransactionsThreshold": 3,
+      "maxHeaviestBranchTipsPerCheckpoint": 10,
+      "randomTipsPerCheckpoint": 3
+    }
   },
   "network": {
     "preferIPv6": false,
--- a/config.json
+++ b/config.json
@@ -83,7 +83,8 @@
     "tipsel": {
       "minHeaviestBranchUnconfirmedTransactionsThreshold": 3,
       "maxHeaviestBranchTipsPerCheckpoint": 10,
-      "randomTipsPerCheckpoint": 3
+      "randomTipsPerCheckpoint": 3,
+      "heaviestBranchSelectionDeadlineMilliseconds": 100
     }
   },
   "network": {
--- a/config.json
+++ b/config.json
@@ -78,8 +78,7 @@
     "merkleTreeFilePath": "coordinator.tree",
     "intervalSeconds": 60,
     "checkpoints": {
-      "maxApproveesCount": 10000,
-      "maxTipsCount": 100
+      "maxTrackedTails": 10000
     },
     "tipsel": {
       "minHeaviestBranchUnconfirmedTransactionsThreshold": 3,
--- a/config.json
+++ b/config.json
@@ -81,9 +81,9 @@
       "maxTrackedTails": 10000
     },
     "tipsel": {
-      "minHeaviestBranchUnconfirmedTransactionsThreshold": 3,
+      "minHeaviestBranchUnconfirmedTransactionsThreshold": 20,
       "maxHeaviestBranchTipsPerCheckpoint": 10,
-      "randomTipsPerCheckpoint": 3,
+      "randomTipsPerCheckpoint": 2,
       "heaviestBranchSelectionDeadlineMilliseconds": 100
     }
   },
--- a/pkg/dag/transaction_root_snapshot_indexes.go
+++ b/pkg/dag/transaction_root_snapshot_indexes.go
@@ -1,0 +1,145 @@
+package dag
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/gohornet/hornet/pkg/model/hornet"
+	"github.com/gohornet/hornet/pkg/model/milestone"
+	"github.com/gohornet/hornet/pkg/model/tangle"
+)
+
+// UpdateOutdatedRootSnapshotIndexes updates the transaction root snapshot indexes of the given transactions.
+// the "outdatedTransactions" should be ordered from latest to oldest to avoid recursion.
+func UpdateOutdatedRootSnapshotIndexes(outdatedTransactions hornet.Hashes, lsmi milestone.Index) {
+	for i := len(outdatedTransactions) - 1; i >= 0; i-- {
+		outdatedTxHash := outdatedTransactions[i]
+
+		cachedTx := tangle.GetCachedTransactionOrNil(outdatedTxHash)
+		if cachedTx == nil {
+			panic(tangle.ErrTransactionNotFound)
+		}
+		GetTransactionRootSnapshotIndexes(cachedTx, lsmi)
+	}
+}
+
+// GetTransactionRootSnapshotIndexes searches the transaction root snapshot indexes for a given transaction.
+func GetTransactionRootSnapshotIndexes(cachedTx *tangle.CachedTransaction, lsmi milestone.Index) (youngestTxRootSnapshotIndex milestone.Index, oldestTxRootSnapshotIndex milestone.Index) {
+	defer cachedTx.Release(true) // tx -1
+
+	// if the tx already contains recent (calculation index matches LSMI)
+	// information about yrtsi and ortsi, return that info
+	yrtsi, ortsi, rtsci := cachedTx.GetMetadata().GetRootSnapshotIndexes()
+	if rtsci == lsmi {
+		return yrtsi, ortsi
+	}
+
+	snapshotInfo := tangle.GetSnapshotInfo()
+
+	youngestTxRootSnapshotIndex = 0
+	oldestTxRootSnapshotIndex = 0
+
+	updateIndexes := func(yrtsi milestone.Index, ortsi milestone.Index) {
+		if (youngestTxRootSnapshotIndex == 0) || (youngestTxRootSnapshotIndex < yrtsi) {
+			youngestTxRootSnapshotIndex = yrtsi
+		}
+		if (oldestTxRootSnapshotIndex == 0) || (oldestTxRootSnapshotIndex > ortsi) {
+			oldestTxRootSnapshotIndex = ortsi
+		}
+	}
+
+	// collect all approvees in the cone that are not confirmed,
+	// are no solid entry points and have no recent calculation index
+	var outdatedTransactions hornet.Hashes
+
+	startTxHash := cachedTx.GetMetadata().GetTxHash()
+
+	// traverse the approvees of this transaction to calculate the root snapshot indexes for this transaction.
+	// this walk will also collect all outdated transactions in the same cone, to update them afterwards.
+	if err := TraverseApprovees(cachedTx.GetMetadata().GetTxHash(),
+		// traversal stops if no more transactions pass the given condition
+		func(cachedTx *tangle.CachedTransaction) (bool, error) { // tx +1
+			defer cachedTx.Release(true) // tx -1
+
+			// first check if the tx was confirmed => update yrtsi and ortsi with the confirmation index
+			if confirmed, at := cachedTx.GetMetadata().GetConfirmed(); confirmed {
+				updateIndexes(at, at)
+				return false, nil
+			}
+
+			if bytes.Equal(startTxHash, cachedTx.GetTransaction().GetTxHash()) {
+				// skip the start transaction, so it doesn't get added to the outdatedTransactions
+				return true, nil
+			}
+
+			// if the tx was not confirmed yet, but already contains recent (calculation index matches LSMI) information
+			// about yrtsi and ortsi, propagate that info
+			yrtsi, ortsi, rtsci := cachedTx.GetMetadata().GetRootSnapshotIndexes()
+			if rtsci == lsmi {
+				updateIndexes(yrtsi, ortsi)
+				return false, nil
+			}
+
+			outdatedTransactions = append(outdatedTransactions, cachedTx.GetTransaction().GetTxHash())
+
+			return true, nil
+		},
+		// consumer
+		func(cachedTx *tangle.CachedTransaction) error { // tx +1
+			defer cachedTx.Release(true) // tx -1
+			return nil
+		},
+		// called on missing approvees
+		func(approveeHash hornet.Hash) error {
+			return fmt.Errorf("missing approvee %v", approveeHash.Trytes())
+		},
+		// called on solid entry points
+		func(txHash hornet.Hash) {
+			updateIndexes(snapshotInfo.EntryPointIndex, snapshotInfo.EntryPointIndex)
+		}, true, false, nil); err != nil {
+		panic(err)
+	}
+
+	// update the outdated root snapshot indexes of all transactions in the cone in order from oldest txs to latest.
+	// this is an efficient way to update the whole cone, because updating from oldest to latest will not be recursive.
+	UpdateOutdatedRootSnapshotIndexes(outdatedTransactions, lsmi)
+
+	// set the new transaction root snapshot indexes in the metadata of the transaction
+	cachedTx.GetMetadata().SetRootSnapshotIndexes(youngestTxRootSnapshotIndex, oldestTxRootSnapshotIndex, lsmi)
+
+	return youngestTxRootSnapshotIndex, oldestTxRootSnapshotIndex
+}
+
+// UpdateTransactionRootSnapshotIndexes updates the transaction root snapshot
+// indexes of the future cone of all given transactions.
+// all the transactions of the newly confirmed cone already have updated transaction root snapshot indexes.
+// we have to walk the future cone, and update the past cone of all transactions that reference an old cone.
+// as a special property, invocations of the yielded function share the same 'already traversed' set to circumvent
+// walking the future cone of the same transactions multiple times.
+func UpdateTransactionRootSnapshotIndexes(txHashes hornet.Hashes, lsmi milestone.Index) {
+	traversed := map[string]struct{}{}
+
+	// we update all transactions in order from oldest to latest
+	for _, txHash := range txHashes {
+
+		if err := TraverseApprovers(txHash,
+			// traversal stops if no more transactions pass the given condition
+			func(cachedTx *tangle.CachedTransaction) (bool, error) { // tx +1
+				defer cachedTx.Release(true) // tx -1
+				_, previouslyTraversed := traversed[string(cachedTx.GetTransaction().GetTxHash())]
+				return !previouslyTraversed, nil
+			},
+			// consumer
+			func(cachedTx *tangle.CachedTransaction) error { // tx +1
+				defer cachedTx.Release(true) // tx -1
+				traversed[string(cachedTx.GetTransaction().GetTxHash())] = struct{}{}
+
+				// updates the transaction root snapshot indexes of the outdated past cone for this transaction
+				GetTransactionRootSnapshotIndexes(cachedTx.Retain(), lsmi) // tx pass +1
+
+				return nil
+			}, true, nil); err != nil {
+			panic(err)
+		}
+	}
+}
--- a/pkg/model/coordinator/coordinator.go
+++ b/pkg/model/coordinator/coordinator.go
@@ -39,10 +39,10 @@ var (
 	ErrNetworkBootstrapped = errors.New("network already bootstrapped")
 )
 
-// coordinatorEvents are the events issued by the coordinator.
-type coordinatorEvents struct {
+// CoordinatorEvents are the events issued by the coordinator.
+type CoordinatorEvents struct {
 	// Fired when a checkpoint transaction is issued.
-	IssuedCheckpoint *events.Event
+	IssuedCheckpointTransaction *events.Event
 	// Fired when a milestone is issued.
 	IssuedMilestone *events.Event
 }
@@ -111,9 +111,9 @@ func New(seed trinary.Hash, securityLvl consts.SecurityLevel, merkleTreeDepth in
 		tipselFunc:              tipselFunc,
 		sendBundleFunc:          sendBundleFunc,
 		milestoneMerkleHashFunc: milestoneMerkleHashFunc,
-		Events: &coordinatorEvents{
-			IssuedCheckpoint: events.NewEvent(CheckpointCaller),
-			IssuedMilestone:  events.NewEvent(MilestoneCaller),
+		Events: &CoordinatorEvents{
+			IssuedCheckpointTransaction: events.NewEvent(CheckpointCaller),
+			IssuedMilestone:             events.NewEvent(MilestoneCaller),
 		},
 	}
 
--- a/pkg/model/coordinator/coordinator.go
+++ b/pkg/model/coordinator/coordinator.go
@@ -58,7 +58,6 @@ type Coordinator struct {
 	minWeightMagnitude      int
 	stateFilePath           string
 	milestoneIntervalSec    int
-	checkpointTransactions  int
 	powFunc                 pow.ProofOfWorkFunc
 	tipselFunc              TipSelectionFunc
 	sendBundleFunc          SendBundleFunc
@@ -100,7 +99,7 @@ func MilestoneMerkleTreeHashFuncWithName(name string) crypto.Hash {
 }
 
 // New creates a new coordinator instance.
-func New(seed trinary.Hash, securityLvl consts.SecurityLevel, merkleTreeDepth int, minWeightMagnitude int, stateFilePath string, milestoneIntervalSec int, checkpointTransactions int, powFunc pow.ProofOfWorkFunc, tipselFunc TipSelectionFunc, sendBundleFunc SendBundleFunc, milestoneMerkleHashFunc crypto.Hash) *Coordinator {
+func New(seed trinary.Hash, securityLvl consts.SecurityLevel, merkleTreeDepth int, minWeightMagnitude int, stateFilePath string, milestoneIntervalSec int, powFunc pow.ProofOfWorkFunc, tipselFunc TipSelectionFunc, sendBundleFunc SendBundleFunc, milestoneMerkleHashFunc crypto.Hash) *Coordinator {
 	result := &Coordinator{
 		seed:                    seed,
 		securityLvl:             securityLvl,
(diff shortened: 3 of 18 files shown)


pkg/model/mselection/heaviest.go
```diff
@@ -66,91 +74,137 @@ func (s *HeaviestSelector) Reset() {
...
+	for i := 0; i < s.maxHeaviestBranchTipsPerCheckpoint; i++ {
+		// run the tip selection for at most 0.1s to keep the view on the tangle recent; this should be plenty
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(100*time.Millisecond))
+
+		tip, count, err := s.selectTip(ctx, tips)
+		if err != nil && err != context.DeadlineExceeded {
+			cancel()
+			break
+		}
+		cancel()
```
alexsporn: Move both `cancel()` before the if?

pkg/model/mselection/heaviest.go
```diff
@@ -340,10 +297,46 @@ func (s *HeaviestSelector) tipItems() []*item {
...
-		panic("empty tips")
+		return nil, ErrNoTipsAvailable
+	}
+
+	index := utils.RandomInsecure(0, len(tips)-1)
+	return &selectedTip{item: tips[index], index: index}, nil
+}
+
+// applyTip set all bits of all referenced transactions of the tip in all existing tips to zero.
+func applyTip(tips []*item, tip *selectedTip) []*item {
```
alexsporn: I feel like the list of items should be an own type with operations on it, instead of helper functions where you pass the list

pkg/model/mselection/heaviest.go
```diff
@@ -191,23 +245,21 @@ func (s *HeaviestSelector) OnNewSolidBundle(bndl *tangle.Bundle) {
 					panic(fmt.Sprintf("transaction not found: %v", hornet.Hash(approveeHash).Trytes()))
 				}
 
-				_, approveeORTSI = s.getTransactionRootSnapshotIndexes(cachedApproveeTx.Retain(), lsmi) // tx +1
+				_, approveeORTSI = dag.GetTransactionRootSnapshotIndexes(cachedApproveeTx.Retain(), lsmi) // tx +1
```
alexsporn: Maybe move this whole check to a separate function where you pass the aproveeHashes and the belowMaxDepth

pkg/model/mselection/heaviest.go
```diff
@@ -191,23 +245,21 @@ func (s *HeaviestSelector) OnNewSolidBundle(bndl *tangle.Bundle) {
...
 			}
 		}
 	}
 
-	// TODO: when len(s.approvers) gets too large trigger a checkpoint to prevent drastic performance hits
-
 	// compute the referenced transactions
 	idx := uint(len(s.approvers))
-	it := &item{hash: bndl.GetTailHash(), index: idx, refs: bitset.New(idx + 1).Set(idx)}
+	it := &item{hash: bndl.GetTailHash(), refs: bitset.New(idx + 1).Set(idx)}
```
alexsporn: I would add a comment here how this works (e.g. we expand the bitset, it references easy added approver, etc.), so we do not have to reverse-engineer this again in a couple of months :-)

alexsporn (approved):

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
 
-	// create an empty list
-	s.tips = list.New()
+// randomTip selects a random tip item from the itemList.
+func (il *itemList) randomTip() (*selectedTip, error) {
+	if len(il.items) == 0 {
+		return nil, ErrNoTipsAvailable
+	}
+
+	index := utils.RandomInsecure(0, len(il.items)-1)
```
karimodm: rand.Intn returns a number [0,n) so the last tip will never be selected.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+	// set all bits of all referenced transactions in all existing tips to zero
+	for _, otherTip := range il.items {
+		otherTip.refs.InPlaceDifference(tip.item.refs)
 	}
+}
 
-	// remove the used tip from the tips list
-	s.removeTip(choosenTip)
+// removeTip removes the tip from the list if items.
+func (il *itemList) removeTip(tip *selectedTip) {
```
karimodm: As noted by FF, this type of removal will change the order of the array. Some functions assume the order of Txs passed to it; make sure this function clearly reflects the fact it does not retain order.

pkg/model/mselection/heaviest.go
```diff
@@ -328,22 +303,62 @@ func (s *HeaviestSelector) removeTip(it *item) {
...
-func randomTip(tips hornet.Hashes) hornet.Hash {
+// randomTipFromTips selects a random tip from the provided slice of tips.
+func randomTipFromTips(tips []*selectedTip) (*selectedTip, error) {
 	if len(tips) == 0 {
-		panic("empty tips")
+		return nil, ErrNoTipsAvailable
 	}
-	return tips[utils.RandomInsecure(0, len(tips)-1)]
+
+	return tips[utils.RandomInsecure(0, len(tips)-1)], nil
```
karimodm: Last tip excluded from rand range.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
-// The selection can be cancelled anytime via the provided context. In this case, it returns the current best solution.
-// selectTip be called concurrently with other HeaviestSelector methods. However, it only considers the tips
-// that were present at the beginning of the selectTip call.
-// TODO: add a proper interface for ms selection that is used by the coordinator
-func (s *HeaviestSelector) selectTip(ctx context.Context) (hornet.Hash, error) {
-	// copy the tips to release the lock at allow faster iteration
-	tips := s.tipItems()
+// selectTip selects a tip to be used for the next checkpoint.
+// it returns a tip, confirming the most transactions in the future cone.
+// the selection can be cancelled anytime via the provided context. in this case, it returns the current best solution.
```
karimodm: Could you add in the comment what the returned 'uint' represents? It is actually used to determine threshold conditions, so good to get documented.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(100*time.Millisecond))
+
+		tip, count, err := s.selectTip(ctx, tipsList)
+		cancel()
+		if err != nil && err != context.DeadlineExceeded {
+			break
+		}
+
+		// if we want to enforce tips, at least collect one
+		if count < uint(s.minHeaviestBranchUnconfirmedTransactionsThreshold) && !(enforceTips && len(result) == 0) {
```
karimodm: If I understand this condition correctly: when you are enforcing tips and you have at least one result -> break out of the heviest tipsel loop, because you may never / soon enough find a tip that meets the `minHeaviestUnconfirmedTransactionsThreshold`.
For me your condition reads: "if I don't meet the threshold, and either I am not enforcing tips or I have at least a result then break".

pkg/model/mselection/heaviest.go
```diff
@@ -328,22 +303,62 @@ func (s *HeaviestSelector) removeTip(it *item) {
...
+}
+
+// GetStats returns the amount of known tips and approvees of s.
+func (s *HeaviestSelector) GetStats() (tipCount int, approveeCount int) {
+	return s.tips.Len(), len(s.approvers)
+}
+
+// checkBelowMaxDepth checks the below max depth criteria for the given approvees.
+// if one of the approvees fails the below max depth criteria, the result is false.
+func checkBelowMaxDepth(approveeHashes map[string]struct{}) bool {
```
karimodm: This check should be implemented as part of URTS as well; can we use one implementation only?

karimodm (changes requested):
Didn't really have a lot of time available to thoroughly understand and comment on this, but I would like to already give some comments.

pkg/model/mselection/heaviest.go
alexsporn: Nope ;-) Look at RandomInsecure again ;-)
We changed it to include the max value, so -1 is correct in this context

pkg/model/mselection/heaviest.go
alexsporn: Nope ;-) Look at RandomInsecure again ;-)
We changed it to include the max value, so -1 is correct in this context

karimodm: Shait, I've looked `RandomInsecure` in master instead of base branch `chrysalis`, my bad ;)

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
 // HeaviestSelector implements the heaviest branch selection strategy.
 type HeaviestSelector struct {
 	sync.Mutex
 
+	minHeaviestBranchUnconfirmedTransactionsThreshold int
+	maxHeaviestBranchTipsPerCheckpoint                int
+	randomTipsPerCheckpoint                           int
+
 	approvers map[trinary.Hash]*item
 	tips      *list.List
```
Wollac:
```suggestion
	approvers map[string]*item // map of all tracked transactions
	tips      *list.List       // list of available tips
```

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
-
 // HeaviestSelector implements the heaviest branch selection strategy.
 type HeaviestSelector struct {
 	sync.Mutex
 
+	minHeaviestBranchUnconfirmedTransactionsThreshold int
+	maxHeaviestBranchTipsPerCheckpoint                int
+	randomTipsPerCheckpoint                           int
+
 	approvers map[trinary.Hash]*item
```
Wollac: approvers is the wrong name as `heaviest` tracks all recent transactions and not only the ones in the future cone of the last MS

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+	tip  *list.Element  // pointer to the element in the tip list
+	refs *bitset.BitSet // BitSet of all the referenced transactions
 }
 
-// New creates a new HeaviestSelector instance.
-func New() *HeaviestSelector {
-	s := &HeaviestSelector{}
-	s.Reset()
-	return s
+type selectedTip struct {
```
Wollac: There is no need to have a dedicated struct for this. It would be more consistent to just return the index and the to the array access outside.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
 }
 
-// ResetCone set all bits of all referenced transactions of the tip in all existing tips to zero.
-func (s *HeaviestSelector) ResetCone(tipHash hornet.Hash) error {
-	s.Lock()
-	defer s.Unlock()
+// applyTip set all bits of all referenced transactions of the tip in all existing tips to zero.
+// this way we can track which parts of the cone would already be referenced by this tip, and
+// correctly calculate the weight of the remaining tips.
+func (il *itemList) applyTip(tip *selectedTip) {
```
Wollac: Apply tip is not a very speaking name, maybe something like reference tip or unset tip could be better.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+			}},
 		count: lastTip.refs.Count(),
 	}
 
 	// loop through all tips and find the one with the most referenced transactions
-	for _, tip := range tips {
+	for index, tip := range tipsList.items {
 		// when the context has been cancelled, return the current best with an error
 		select {
 		case <-ctx.Done():
```
Wollac: I am not sure whether in this particular implementation `context` makes sense. This is just O(n) and the channel is checked every tip which causes quite a bit of overhead.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
-			return randomTip(best.tips), ctx.Err()
+			selected, err := randomTipFromTips(best.tips)
```
Wollac: `best` was initialized explicitly in a way, that it _always_ contains at least one tip. So, there is no need to have this error handling here and the code could be simplified.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+			best.tips = append(best.tips, &selectedTip{
+				item:  tip,
+				index: index,
+			})
 		}
 	}
 
-	// TODO: is it really to select a random tip? Maybe prefer the older (or younger) tip instead
-	return randomTip(best.tips), nil
+	selected, err := randomTipFromTips(best.tips)
```
Wollac: again, there will never be an error

pkg/model/mselection/heaviest.go
```diff
@@ -66,91 +74,137 @@ func (s *HeaviestSelector) Reset() {
...
+		return nil, 0, err
+	}
+	return selected, best.count, nil
 }
 
-// SelectTip selects a tip to be used for the next milestone.
-func (s *HeaviestSelector) SelectTip() (hornet.Hash, error) {
+// SelectTips tries to collect tips that confirm the most transactions in the future cone.
+// best tips are determined by counting the referenced transactions (heaviest branches) and by "removing" the
+// transactions of the referenced cone of the already choosen tips in the bitsets of the available tips.
```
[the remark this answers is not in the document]
Wollac: The dog is right! this should be corrected

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+	selected, err := randomTipFromTips(best.tips)
+	if err != nil {
+		return nil, 0, err
+	}
+	return selected, best.count, nil
 }
 
-// SelectTip selects a tip to be used for the next milestone.
-func (s *HeaviestSelector) SelectTip() (hornet.Hash, error) {
+// SelectTips tries to collect tips that confirm the most transactions in the future cone.
```
Wollac: Future cone of what? We are just keeping track of the k most recent transactions (that are not below max depth). So references and approves are really be considered, it should just be called recent transactions (since the last `reset`)

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
 
-	// run the tip selection for at most 0.1s to keep the view on the tangle recent; this should be plenty
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(100*time.Millisecond))
-	defer cancel()
-	return s.selectTip(ctx)
+	var result hornet.Hashes
+
+	for i := 0; i < s.maxHeaviestBranchTipsPerCheckpoint; i++ {
+		// run the tip selection for at most 0.1s to keep the view on the tangle recent; this should be plenty
+		ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(100*time.Millisecond))
```
Wollac: The context here is not good:
- the cancel should be deferred to guarantee cleanup also on panics.
- the context should come from "outside" or the deadline should be configurable.
- if at all a local context is used, it should be outside of the for loop to have a deadline and guarantee that the select tip *always* returns within the deadline

pkg/model/mselection/heaviest.go
Wollac: The condition is indeed very hard to understand.
Consider replacing `enforceTips` with `minRequiredTips int` or something similar. Even though this would always only ever be `0` or `1`. it makes it much easier to understand what this parameter is supposed to do and also the condition would get considerably easier.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+	s.reset()
+
+	return result, nil
 }
 
 // OnNewSolidBundle adds a new bundle to be processed by s.
 // The bundle must be solid and OnNewSolidBundle must be called in the order of solidification.
 // We also have to check if the bundle is below max depth.
-func (s *HeaviestSelector) OnNewSolidBundle(bndl *tangle.Bundle) {
+func (s *HeaviestSelector) OnNewSolidBundle(bndl *tangle.Bundle) (tipCount int, approveeCount int) {
```
Wollac: `approveeCount` is the wrong name. It just returns the number of solid bundles (not be low max depth) since the last reset, independently of what they are approving

pkg/model/mselection/heaviest.go
```diff
@@ -328,22 +303,62 @@ func (s *HeaviestSelector) removeTip(it *item) {
 	it.tip = nil
 }
 
-// tipItems returns a copy of the items corresponding to tips.
-func (s *HeaviestSelector) tipItems() []*item {
+// copyTipItemsToList returns a copy of the items corresponding to tips.
+func (s *HeaviestSelector) copyTipItemsToList() *itemList {
```
Wollac: The name copy is misleading as this is just a flat copy of pointers. If something modifies the items we would get race-conditions.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+// selecting at least one tip can be enforced, even if none of the heaviest branches matches the
+// "minHeaviestBranchUnconfirmedTransactionsThreshold" criteria with "enforceTips".
+// if at least one heaviest branch tip was found, "randomTipsPerCheckpoint" random tips are added
+// to add some additional randomness to prevent parasite chain attacks.
+func (s *HeaviestSelector) SelectTips(enforceTips bool) (hornet.Hashes, error) {
+
+	// copy the tips to release the lock to allow faster iteration
+	// and to get a frozen view of the tangle, so an attacker can't
+	// create heavier branches while we are searching the best tips
+	tipsList := s.copyTipItemsToList()
```
Wollac: This only works without race-conditions when the items are immutable! Currently this seems to be the case but nothing is actually enforcing this, so at least there should be a comment mentioning this.

config.json
```diff
@@ -77,7 +77,15 @@
     "stateFilePath": "coordinator.state",
     "merkleTreeFilePath": "coordinator.tree",
     "intervalSeconds": 60,
-    "checkpointTransactions": 5
+    "checkpoints": {
+      "maxApproveesCount": 10000,
```
Wollac: I really don't like the name approvee for this parameter. Effectively, it just limits the number of recent transactions the (heaviest) tip selector keeps track of (in its bitset). I think a name in that direction would be clearer.

plugins/coordinator/plugin.go
```diff
@@ -99,42 +118,68 @@ func initCoordinator(bootstrap bool, startIndex uint32) (*coordinator.Coordinato
...
+		if criticalErr := coo.Bootstrap(); criticalErr != nil {
+			log.Panic(criticalErr)
+		}
+
+	coordinatorLoop:
+		for {
+			select {
+			case <-nextCheckpointSignal:
+				// check the thresholds again, because a new milestone could have been issued in the meantime
+				if tipCount, approveeCount := selector.GetStats(); (tipCount < maxTipsCount) && (approveeCount < maxApproveesCount) {
```
Wollac: Currently, a checkpoint is issued when the bitsets gets too large - this is absolutely crucial - and when the number of tips gets too large.
I am not so sure about the tip count, the heaviest tip selection is so fast that it would work for hundred thousands of tips. So what exactly is the motivation behind the tip limit? For performance it would make much more sense to limit the actual total runtime (like in the original version) instead of the number of tips. Also, if an attacker spams blow balls this would trigger many checkpoints which is actually not great.
If the purpose of that parameter is to trigger more checkpoints in very high TPS scenarios it would be favorable to use the weight of the heaviest tip as a threshold. This could be returned by the selector without computation.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
+// transactions of the referenced cone of the already choosen tips in the bitsets of the available tips.
+// only tips are considered that were present at the beginning of the SelectTips call,
+// to prevent attackers from creating heavier branches while we are searching the best tips.
+// "maxHeaviestBranchTipsPerCheckpoint" is the amount of tips that are collected if
+// the current best tip is not below "UnconfirmedTransactionsThreshold" before.
+// selecting at least one tip can be enforced, even if none of the heaviest branches matches the
+// "minHeaviestBranchUnconfirmedTransactionsThreshold" criteria with "enforceTips".
+// if at least one heaviest branch tip was found, "randomTipsPerCheckpoint" random tips are added
+// to add some additional randomness to prevent parasite chain attacks.
+func (s *HeaviestSelector) SelectTips(enforceTips bool) (hornet.Hashes, error) {
```
Wollac: The `SelectTips` is actual a greedy heuristic algorithm. It iteratively selects the n "heaviest" tip, removes its past cone and then continues with the now heaviest tip until the required number of tips is reached. This will not be the optimal solution.
The "heaviest pair selection" actually does the same for n=2, but it selects the _optimal_ two tips which can be considerably stronger than the first two tips returned by this. (This could also be extended for n=3, etc to find the optimal solution in O(n^3).)
I would still be very very much interested to see how this behaves. It would be my guess, that the the quality is actually very much similar but it would work like the current protocol with only one Milestone bundle which could be preferable.

pkg/model/coordinator/coordinator.go
```diff
@@ -210,35 +209,53 @@ func (coo *Coordinator) InitState(bootstrap bool, startIndex milestone.Index) er
...
+		lastCheckpointHash = coo.lastMilestoneHash
+	} else {
+		// reference the last checkpoint
+		lastCheckpointHash = coo.lastCheckpointHash
 	}
 
-	coo.lastCheckpointCount++
-	lastCheckpointHash := hornet.HashFromHashTrytes(b[0].Hash)
-	coo.lastCheckpointHash = &lastCheckpointHash
+	for i, tip := range tips {
```
Wollac, line 242: This issues n checkpoints for n tips. In a tree the same result could be achieved with only log(n) checkpoints. Especially since the checkpoint require a considerable amount of PoW this should be considered!

pkg/model/mselection/heaviest.go
```diff
@@ -176,38 +268,16 @@ func (s *HeaviestSelector) OnNewSolidBundle(bndl *tangle.Bundle) {
...
-				cachedApproveeTx.Release(true)
-			}
-
-			// if the confirmationIdx to LSMI delta of the approvee is equal or greater belowMaxDepth, the tip is invalid.
-			// "equal" is important because the next milestone would reference this transaction.
-			if lsmi-approveeORTSI >= belowMaxDepth {
-				return
-			}
-		}
+	if !checkBelowMaxDepth(approveeHashes) {
```
Wollac: What is generally, the current philosophy with BMD? Should transactions that became solid, but are below max depth really trigger this event? Or could they be filtered out earlier? For what would BMD transactions be relevant are they not just invalid and be discarded?
Not having to check for below max depth would also make the tip selector code much cleaner as the BMD check is actually something not really related to this particular tip selection.

plugins/coordinator/plugin.go
```diff
@@ -99,42 +118,68 @@ func initCoordinator(bootstrap bool, startIndex uint32) (*coordinator.Coordinato
...
+					continue
+				}
+
+				// issue a checkpoint
+				if err := coo.IssueCheckpoint(); err != nil {
```
Wollac, line 158: Can we still integrate the "traditional" checkpoints to limit attackers with a very high throughput? This could still be a helpful mechanic.

pkg/model/mselection/heaviest.go
```diff
@@ -20,137 +19,230 @@ import (
...
 
+	minHeaviestBranchUnconfirmedTransactionsThreshold int
+	maxHeaviestBranchTipsPerCheckpoint                int
+	randomTipsPerCheckpoint                           int
+
 	approvers map[trinary.Hash]*item
 	tips      *list.List
 }
 
 type item struct {
```
Wollac: consider renaming `item` to `bundle` or similar.

plugins/coordinator/plugin.go
muXxer: So reintroduce the root transaction thing? Or what do you mean exactly? The whole reason to remove it was because of bad confirmation rates in my simulation.

pkg/dag/transaction_root_snapshot_indexes.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// collect all approvees in the cone that are not confirmed,
+	// are no solid entry points and have no recent calculation index
+	var outdatedTransactions hornet.Hashes
+
+	startTxHash := cachedTx.GetMetadata().GetTxHash()
+
+	// traverse the approvees of this transaction to calculate the root snapshot indexes for this transaction.
+	// this walk will also collect all outdated transactions in the same cone, to update them afterwards.
+	if err := TraverseApprovees(cachedTx.GetMetadata().GetTxHash(),
```
GalRogozinski, line 59: Is there protection against stitching attack DOS?

Explanation:
Someone starts to create a huge parasitic subtangle offline.
Then one publishes it, then one solidifies it by broadcasting a missing hole.

My fear is that we can be stuck in I/O with long traversals

I am not sure though that such an attack will work since:
1. it depends on solidification events
2. maybe caching relieves it?

Well.. I am only reading code atm and not testing yet...

pkg/model/coordinator/coordinator.go
```diff
@@ -210,35 +209,53 @@ func (coo *Coordinator) InitState(bootstrap bool, startIndex milestone.Index) er
...
+		b, err := createCheckpoint(tip, lastCheckpointHash, coo.minWeightMagnitude, coo.powFunc)
```
GalRogozinski, line 243: Since you are using `finalizeInsecure` underneath, can then an attacker recognize checkpoints by looking for bundles with `M`?

pkg/model/mselection/heaviest.go
```diff
@@ -20,330 +18,303 @@ import (
...
-			best.tips = hornet.Hashes{tip.hash}
-			best.count = c
-		} else if c == best.count {
-			best.tips = append(best.tips, tip.hash)
+		tip, count, err := s.selectTip(tipsList)
+		if err != nil {
+			break
 		}
+
+		if (len(result) > minRequiredTips) && ((count < uint(s.minHeaviestBranchUnconfirmedTransactionsThreshold)) || deadlineExceeded) {
```
GalRogozinski, line 209: This condition still confuses me
let say count is larger than `minHeaviestBranchUnconfirmedTransactionsThreshold`

why don't I break?

muXxer: We only break if it is lower, that's the whole point

pkg/model/coordinator/coordinator.go
muXxer: What can an attacker do with knowing the checkpoint transaction?
A checkpoint is not the same like before, we don't check the future cone of the checkpoints anymore.

pkg/model/mselection/heaviest.go
GalRogozinski: I get it now

Btw, for pure readability I would change to `>= minRequiredTips` and then do +1 to the values you pass
This is just because `minRequiredTips` makes it sound like there should be `=`

pkg/model/coordinator/coordinator.go
GalRogozinski: I see now
So there's no parasite chain protection at all anymore and just complete trust that URTS will help merge everything in

I am kinda sad to see that we don't have it but maybe tests will prove that it is actually not so important

PR #572, opened 2020-07-22
Added whiteflag test setup

alexsporn:
- Added the necessary helper functions and setup to create and visualize whiteflag confirmation
- Fixed a bug when calculating the address mutations during whiteflag confirmation


pkg/whiteflag/test/whiteflag_test.go
```diff
@@ -0,0 +1,599 @@
...
+	for _, value := range balances {
+		sum += value
+	}
+	// Move remaining supply to 999..999
+	balances[string(hornet.NullHashBytes)] = consts.TotalSupply - sum
+
+	store := mapdb.NewMapDB()
+	setupTestEnvironment(t, store)
+
+	solidEntryPointsOnce.Do(func() {
```
muXxer: Why is it called `solidEntryPointsOnce` ?

pkg/whiteflag/test/whiteflag_test.go
```diff
@@ -0,0 +1,599 @@
...
+
+func TestWhiteFlagWithMultipleConflicting(t *testing.T) {
+
+	// Fill up the balances
+	balances := make(map[string]uint64)
+	balances[string(generateAddress(t, seed1, 0))] = 1000
+
+	milestones := setupCoordinatorAndIssueInitialMilestones(t, balances, 3)
+
+	// Issue some transactions
```
muXxer, line 565: Could you please add some comments here? It's hard to read the testcase :)
For example what would be the final balance you would expect on which seed, or something like that.

alexsporn: Do you mean as a comment of every bundle? At the end I check every seed

pkg/whiteflag/test/whiteflag_test.go
alexsporn: forgot to rename it at the end, but actually it protects that we only itinialize the solid entry points once. We have some global state there that I cannot reset betweeen each test. I'd need to adapt the code, but we made it so to actually catch wrong usage. I'll rename the once

pkg/whiteflag/test/whiteflag_test.go
alexsporn: OK, I see that without labels it looks difficult to read. My IDE was showing it to me like this:
Screenshot 2020-07-24 at 09 08 23

muXxer (approved):

PR #582, opened 2020-07-30
Ports and refactors the integration framework from GoShimmer

luca-moser:
Brings an integration framework which uses Docker to spawn networks of Hornet nodes in order to assert behavior on them in an automated way.

This is a port from GoShimmer but was heavily refactored for additional features:
* Static peering which decreases setup time immensely
* The ability to define own static peering patterns in form of maps.
* Profiling of heap, CPU etc. through function calls on nodes
* Graphing of TPS, memory, conf. rate etc. in an output HTML file
* Configuration of nodes can be overwritten before they are spawned so one is able to for example let one node in a network run with a different configuration.

The PR ships with 3 tests:
* `common` which is just a sanity check to see whether a static network of HORNET nodes sync up.
* `autopeering` the same as `common` but with autopeering.
* `benchmark` which for 30 seconds instructs a network of 4 nodes to spam zero-val txs and produces CPU profiles and TPS/memory charts.

Networks spawned always include a Coordinator and every node is initialized with a snapshot file with a genesis address holding all tokens. Tests which require other setups may easily add their own initial states by putting assets in the `integration-tests/assets` folder which is available in every spawned node.

The integration tests run with every "PR open/Push" event. The GitHub workflow folder has been extended with the given file. Unfortunately there is no easy way right now to share the built Docker image between jobs, so each job (which runs in parallel) first builds the dev image of the code. This might be changed/improved in the future.


.github/workflows/test_HORNET.yml
```diff
@@ -28,4 +28,4 @@ jobs:
         with:
           timeout_minutes: 5
           max_attempts: 3
-          command: go test ./...
+          command: go test `go list ./... | grep -v integration-tests`
```
muXxer, line 31: Will that prevent other tests to run that are not inside the integration-tests folder?

config.json
```diff
@@ -97,7 +97,9 @@
     "alias": "",
     "showAliasInGetNodeInfo": false,
     "disablePlugins": [],
-    "enablePlugins": []
+    "enablePlugins": [
+      "Coordinator"
```
muXxer: Pls remove me! :)

integration-tests/runTests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+TEST_NAMES='common'
```
muXxer: Is the file `runTests.sh` only for a local test?
If not, should the other tests also be added here?

pkg/peering/reconnect.go
```diff
@@ -95,6 +95,7 @@ next:
 // adds the given peers to the reconnect pool.
 func (m *Manager) moveInitialPeersToReconnectPool(peers []*config.PeerConfig) {
 	for _, peerConf := range peers {
+		fmt.Println("conf", peerConf)
```
muXxer: This can be removed I guess.

integration-tests/tester/framework/autopeered_network.go
```diff
@@ -0,0 +1,295 @@
...
+	// create pumba container for every peer in the partition
+	pumbas := make([]*DockerContainer, len(peers))
+	for i, p := range peers {
+		name := partitionName + p.Name + containerNameSuffixPumba
+		pumba, err := n.createPumba(name, p.Name, targetIPs)
+		if err != nil {
+			return nil, err
+		}
+		pumbas[i] = pumba
+		time.Sleep(1 * time.Second)
```
muXxer, line 199: why do we have to wait here?

integration-tests/tester/framework/autopeered_network.go
```diff
@@ -0,0 +1,295 @@
...
+// createPartition creates a partition with the given peers.
+// It starts a Pumba container for every peer that blocks traffic to all other partitions.
+func (n *AutopeeredNetwork) createPartition(peers []*Node) (*Partition, error) {
+	peersMap := make(map[string]*Node)
+	for _, peer := range peers {
+		peersMap[peer.ID().String()] = peer
+	}
+
+	// block all traffic to all other peers except in the current partition
+	var targetIPs []string
```
muXxer, line 180: the targetIPs are the IPs that get blocked by a "pumba" container?
Should we maybe rename them to "blockedIPs" ?

integration-tests/tester/framework/static_network.go
```diff
@@ -0,0 +1,134 @@
...
+	return nil
+}
+
+// AwaitPeering awaits until all nodes are peered according to the peering layout.
+func (n *StaticNetwork) AwaitPeering(ctx context.Context) error {
+	log.Println("verifying peering...")
+	for nodeID, layoutNeighbors := range n.layout {
+		node := n.Nodes[nodeID]
+		for {
+			if err := returnErrIfCtxDone(ctx, ErrNodesDidNotPeerInTime); err != nil {
```
muXxer, line 109: Wouldn't that loop consume 100% CPU? Should we add a small time.sleep here?

integration-tests/tester/framework/node.go
```diff
@@ -0,0 +1,148 @@
...
+
+				if _, err = p.WebAPI.BroadcastTransactions(toSend...); err != nil {
+					return
+				}
+
+				atomic.AddInt32(&spammed, int32(batch))
+			}
+		}()
+	}
+	return spammed, nil
```
muXxer, line 148: Shouldn't you wait for the waitgroup to be done?

integration-tests/tester/framework/network.go
```diff
@@ -0,0 +1,260 @@
...
+func (n *Network) PrefixName(suffix string) string {
+	return fmt.Sprintf("%s-%s", n.Name, suffix)
+}
+
+// AwaitOnline awaits until all nodes are online or the given context is canceled.
+func (n *Network) AwaitOnline(ctx context.Context) error {
+	log.Println("waiting for nodes to become online...")
+	for _, node := range n.Nodes {
+		for {
+			if err := returnErrIfCtxDone(ctx, ErrNodesNotOnlineInTime); err != nil {
```
muXxer, line 52: Small time sleep here?

integration-tests/tester/framework/network.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return nil
+}
+
+// AwaitAllSync awaits until all nodes see themselves as synced.
+func (n *Network) AwaitAllSync(ctx context.Context) error {
+	log.Println("waiting for nodes to become synced...")
+	for _, node := range n.Nodes {
+		for {
+			if err := returnErrIfCtxDone(ctx, ErrNodesDidNotSyncInTime); err != nil {
```
muXxer, line 69: Small time sleep here?

muXxer (changes requested):
Very cool framework!

.github/workflows/test_HORNET.yml
luca-moser: This command does exactly the opposite, note the `-v` flag on `grep` excluding everything with `integration-tests`. This command will run all tests except integration-tests.

muXxer: ahh I thought that was "verbose" or something, lol. thx

integration-tests/runTests.sh
luca-moser: `runTests.sh` is only meant to run integration tests locally since on the CI every integration test is running in parallel. I'll put back the other tests.

pkg/peering/reconnect.go
luca-moser: Removed

integration-tests/tester/framework/autopeered_network.go
luca-moser: This is just an additional grace period after starting each container. While we are actually waiting 5 seconds after all of them have been spawned I believe it is still a good idea to give each separate container such grace time.

integration-tests/tester/framework/autopeered_network.go
luca-moser: Lets not rename it because pumba is not only for blocking traffic.

integration-tests/tester/framework/static_network.go
luca-moser: At worst it would maybe consume 100% of one core if neighbors are completely offline.

integration-tests/tester/framework/node.go
luca-moser: Good catch.

integration-tests/tester/framework/network.go
luca-moser: At worst one core would spin up to 100%. I won't put a sleep.

integration-tests/tester/framework/network.go
luca-moser: At worst one core would spin up to 100%. I won't put a sleep.

muXxer (approved):
LGTM

PR #583, opened 2020-07-31
Use dag helper in whiteflag

muXxer:
This refactoring was done for better readability of whiteflag.

### DIFF
--- a/pkg/dag/helper.go
+++ b/pkg/dag/helper.go
@@ -3,6 +3,7 @@ package dag
 import (
 	"bytes"
 	"container/list"
+	"fmt"
 
 	"github.com/pkg/errors"
 
@@ -51,7 +52,7 @@ func FindAllTails(startTxHash hornet.Hash, skipStartTx bool, forceRelease bool) 
 		// called on solid entry points
 		func(approveeHash hornet.Hash) {
 			// Ignore solid entry points (snapshot milestone included)
-		}, true, false, nil)
+		}, true, false, false, nil)
 
 	return tails, err
 }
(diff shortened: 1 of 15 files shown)


pkg/dag/helper.go
```diff
@@ -139,8 +140,24 @@ func processStackApprovees(stack *list.List, processed map[string]struct{}, chec
...
+		// load up bundle to retrieve trunk and branch of the head tx
+		cachedBundle := tangle.GetCachedBundleOrNil(currentTxHash)
+		if cachedBundle == nil {
+			return fmt.Errorf("%w: bundle %s of candidate tx %s doesn't exist", tangle.ErrBundleNotFound, cachedTx.GetTransaction().Tx.Bundle, currentTxHash.Trytes())
+		}
+		defer cachedBundle.Release(true)
+
+		cachedBundleHeadTx := cachedBundle.GetBundle().GetHead()
+		defer cachedBundleHeadTx.Release(true)
+		trunkHash = cachedBundleHeadTx.GetTransaction().GetTrunkHash()
```
alexsporn: cachedBundle.GetTrunk(true)
cachedBundle.GetBranch(true)

pkg/whiteflag/confirmation.go
```diff
@@ -28,21 +28,25 @@ type ConfirmedMilestoneStats struct {
...
 
 	ts := time.Now()
-	confirmation, err := ComputeConfirmation(tangle.GetMilestoneMerkleHashFunc(), cachedMsBundle.Retain(), milestoneIndex)
+
+	confirmation, err := ComputeConfirmation(tangle.GetMilestoneMerkleHashFunc(), msBundle.GetTailHash())
 	if err != nil {
 		// According to the RFC we should panic if we encounter any invalid bundles during confirmation
 		return nil, fmt.Errorf("confirmMilestone: whiteflag.ComputeConfirmation failed with Error: %v", err)
 	}
+	confirmation.MilestoneIndex = milestoneIndex
```
alexsporn: Don't like that we set these values in an object that was created in ComputeConfirmation

muXxer: But the COO doesn't know about its final tail hash before the bundle gets issued... So if we want to reuse the code... what would be the best practice?

alexsporn: This method returns an object not containing it, and the confirmation.go creates an object with this + index + hash?

alexsporn (approved):

PR #589, opened 2020-08-04
Added optional PoW method

IoTMOD:


pkg/pow/pow.go
```diff
@@ -0,0 +1,113 @@
...
+const powsrvReinitCooldown = 30 * time.Second
+
+// Handler struct
+type Handler struct {
+	sync.RWMutex
+
+	powsrvClient      *powsrvio.PowClient
+	localPoWFunc      pow.ProofOfWorkFunc
+	powType           string
+	lastInitTimestamp int64
```
muXxer: Use time.Time

pkg/pow/pow.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// tryReinitPowsrvWithoutLocking tries to reinit powsrv.io if the connection got lost.
+// the lock needs to be handled by the caller
+func (h *Handler) tryReinitPowsrvWithoutLocking() {
+	if h.powsrvClient == nil {
+		return
+	}
+
+	if h.lastInitTimestamp+powsrvReinitCooldown.Nanoseconds() <= time.Now().UnixNano() {
```
muXxer: Use time.Since

pkg/pow/pow.go
```diff
@@ -0,0 +1,113 @@
...
+		h.powsrvClient.Close()
+		h.powsrvClient.Init()
+		h.lastInitTimestamp = time.Now().UnixNano()
+	}
+}
+
+// NewHandler creates a new PoW handler instance
+func NewHandler() (*Handler, error) {
```
muXxer: Call it "New"

pkg/pow/pow.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var localPoWFunc pow.ProofOfWorkFunc
```
muXxer: Create a pow plugin and init the Handler globally with:

```go
var (
	PLUGIN      = node.NewPlugin("PoW", node.Enabled, configure, run)
	log         *logger.Logger
	handler     *pow.Handler
	handlerOnce sync.Once
)


// Handler gets the pow Handler instance.
func Handler() *pow.Handler {
	handlerOnce.Do(func() {
		// init the pow handler with all possible settings
	})
	return handler
}
```

Then you can call `pow.Handler()` from all plugins that use it, and you can connect the shutdown event cleanly with its own priority.

pkg/pow/pow.go
muXxer: Why return an error if it can never happen?

pkg/pow/pow.go
```diff
@@ -0,0 +1,113 @@
...
+
+// GetPoWType returns the fastest available PoW type which gets used for PoW requests
+func (h *Handler) GetPoWType() string {
+	return h.powType
+}
+
+// DoPoW calculates the PoW
+// Either with the fastest available local PoW function or with the help of powsrv.io (optional, POWSRV_API_KEY env var must be available)
+func (h *Handler) DoPoW(trytes trinary.Trytes, mwm int, parallelism ...int) (nonce string, err error) {
+	h.Lock()
```
muXxer: Only one PoW request at a time would be possible with this.
Do we need a lock here?

plugins/coordinator/plugin.go
```diff
@@ -68,15 +69,20 @@ func configure(plugin *node.Plugin) {
 	tangleplugin.SetUpdateSyncedAtStartup(true)
 
 	var err error
-	coo, err = initCoordinator(*bootstrap, *startIndex)
+	powHandler, err = pow.NewHandler()
```
muXxer: Use the proposed global PoW handler from the plugin.

plugins/spammer/plugin.go
```diff
@@ -68,6 +70,13 @@ func configure(plugin *node.Plugin) {
 	checkPeersConnected = node.IsSkipped(coordinator.PLUGIN)
 	spammerAvgHeap = utils.NewTimeHeap()
 
+	var err error
+	if powHandler, err = pow.NewHandler(); err != nil {
```
muXxer: Use the proposed global PoW handler from the plugin.

plugins/spammer/plugin.go
```diff
@@ -132,6 +141,14 @@ func run(_ *node.Plugin) {
 		return
 	}
 
+	// close the PoW handler on shutdown
+	daemon.BackgroundWorker("Close PoW handler", func(shutdownSignal <-chan struct{}) {
```
muXxer: This can be moved to the global PoW handler plugin.

plugins/webapi/pow.go
```diff
@@ -53,6 +52,15 @@ func attachToTangle(i interface{}, c *gin.Context, _ <-chan struct{}) {
 		return
 	}
 
+	// Init the PoW handler
+	powHandler, err := pow.NewHandler()
```
muXxer: Use the proposed global PoW handler from the plugin.

muXxer (changes requested):

pkg/whiteflag/test/whiteflag_test.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/gohornet/hornet/pkg/profile"
 	"github.com/gohornet/hornet/pkg/utils"
 	"github.com/gohornet/hornet/pkg/whiteflag"
+	hornet_pow "github.com/gohornet/hornet/plugins/pow"
```
alexsporn: plugin in a package? 🙈

muXxer: Fixed :carlos:

alexsporn (approved):

muXxer (approved):

PR #593, opened 2020-08-06
Fix revalidation/pruning/CTPS calculation/neighboring + add extended Heartbeat message

muXxer:


pkg/whiteflag/confirmation.go
```diff
@@ -132,13 +132,15 @@ func ConfirmMilestone(cachedMsBundle *tangle.CachedBundle, forEachConfirmedTx fu
...
 	for _, txHash := range mutations.TailsIncluded {
 		if err := forEachBundleTxWithTailTxHash(txHash, func(tx *tangle.CachedTransaction) {
-			tx.GetMetadata().SetConfirmed(true, milestoneIndex)
-			tx.GetMetadata().SetRootSnapshotIndexes(milestoneIndex, milestoneIndex, milestoneIndex)
-			conf.TxsConfirmed++
-			conf.TxsValue++
-			metrics.SharedServerMetrics.ValueTransactions.Inc()
-			metrics.SharedServerMetrics.ConfirmedTransactions.Inc()
-			forEachConfirmedTx(tx, milestoneIndex, confirmationTime)
+			if !tx.GetMetadata().IsConfirmed() {
```
alexsporn, line 135: why would this be true?

plugins/snapshot/pruning.go
```diff
@@ -44,15 +44,17 @@ loopOverUnconfirmed:
...
 			if bundles != nil && len(bundles) > 0 {
 				for _, bndl := range bundles {
 					if bndl.GetBundle().IsMilestone() {
 						cachedTx.Release(true) // tx -1
+						bundles.Release(true)  // bundles -1
 						// Do not prune unconfirmed tx that are part of a milestone
 						continue loopOverUnconfirmed
 					}
 				}
+				bundles.Release(true) // bundles -1
```
alexsporn, line 57: Oops 🙈 mea culpa

pkg/whiteflag/confirmation.go
alexsporn: Is it due to the funny bundle tx reuse from discord?

plugins/snapshot/pruning.go
muXxer: I also missed that in the review :)

pkg/whiteflag/confirmation.go
muXxer: Because of f*cking bundles.... and their tail reattachments, you would count all the single txs of this bundle several times in the confirmation.

alexsporn (approved):

PR #611, opened 2020-08-12
Use cached metadata instead tx to reduce disc io

muXxer:


pkg/model/tangle/transaction_storage.go
```diff
@@ -150,19 +168,46 @@ func GetCachedTransactionOrNil(txHash hornet.Hash) *CachedTransaction {
...
 		return nil
 	}
-	return storedTx.(*hornet.Transaction)
+
+	addTrunkBranchHashesToMetadata(cachedMeta.Retain())
+
+	return &CachedMetadata{CachedObject: cachedMeta}
+}
+
+func addTrunkBranchHashesToMetadata(cachedMetadata objectstorage.CachedObject) {
```
alexsporn: Rename since now it does more than that?

alexsporn (approved):

pkg/model/tangle/transaction_storage.go
```diff
@@ -60,10 +66,18 @@ func (c *CachedTransaction) GetTransaction() *hornet.Transaction {
 	return c.tx.Get().(*hornet.Transaction)
 }
 
+func (c *CachedTransaction) GetCachedMetadata() *CachedMetadata {
+	return &CachedMetadata{c.metadata}
```
alexsporn: this should probably retain the metadata

plugins/snapshot/pruning.go
```diff
@@ -111,13 +111,13 @@ func pruneTransactions(txsToCheckMap map[string]struct{}) int {
 		cachedTx.Release(true) // tx -1
```
alexsporn: Release at the end of the loop, or maybe do a Consume?

plugins/tangle/solidifier.go
```diff
@@ -56,22 +56,26 @@ func TriggerSolidifier() {
...
+func markTransactionAsSolid(cachedTxMeta *tangle.CachedMetadata) {
 	// Construct the complete bundle if the tail got solid (before setting solid flag => otherwise not threadsafe)
-	if cachedTx.GetTransaction().IsTail() {
-		tangle.OnTailTransactionSolid(cachedTx.Retain()) // tx pass +1
+	if cachedTxMeta.GetMetadata().IsTail() {
+		cachedTx := tangle.GetCachedTransactionOrNil(cachedTxMeta.GetMetadata().GetTxHash())
+		if cachedTx == nil {
+			log.Panicf("markTransactionAsSolid: Transaction not found: %v", cachedTxMeta.GetMetadata().GetTxHash().Trytes())
+		}
+		tangle.OnTailTransactionSolid(cachedTx) // tx pass +1
```
alexsporn, line 66: OnTailTransactionSolid and then tryConstructBundle can use metadata only too as far as I can see

alexsporn: hmm... tryConstructBundle is also using CurrentIndex and LastIndex, so maybe not :(

muXxer: No we need the Transaction indexes to construct the bundle, which are only stored in the transaction.
Was thinking about adding that to the metadata as well, but this function is only called when we receive a transaction, so we have the object in cache anyway.... and we will remove the bundle scam soom.

plugins/webapi/debug.go
```diff
@@ -213,41 +213,41 @@ func searchEntryPoints(i interface{}, c *gin.Context, _ <-chan struct{}) {
...
 				}
 			}
 
 			return true, nil
 		},
 		// consumer
-		func(cachedTx *tangle.CachedTransaction) error { // tx +1
-			defer cachedTx.Release(true) // tx -1
+		func(cachedTxMeta *tangle.CachedMetadata) error { // meta +1
+			defer cachedTxMeta.Release(true) // meta -1
```
alexsporn: maybe use Consume?

alexsporn (approved):

alexsporn (approved):

PR #616, opened 2020-08-15
feature(dashboard): Show database size in GB if size > 1 GB

rajivshah3:
# Description

This PR changes the database size graph so that it shows the size in GB if size > 1 GB. This makes it a bit easier to read compared to MB. 

- Conversion from bytes to MB was removed from `NodeStore`
- New `dbSizeChartOpts` was created to represent size in GB
- The graph will switch from `dbSizeChartOptsMB` (axis and tooltip labels shown in MB) to `dbSizeChartOptsGB`when database size exceeds 1 GB

One consequence of this change is that the spent addresses and snapshot may show as 0.00 GB:
image

I can show sizes in GB with 3 decimal places (instead of 2) if desired

## Type of change

# How Has This Been Tested?

Tested manually

# Checklist:

- [ ] I have selected the `develop` branch as the target branch

### DIFF
--- a/plugins/dashboard/frontend/src/app/components/Misc.tsx
+++ b/plugins/dashboard/frontend/src/app/components/Misc.tsx
@@ -15,6 +15,11 @@ interface Props {
     nodeStore?: NodeStore;
 }
 
+// 1 MB in bytes
+const MB_TO_BYTES = 1024 * 1024;
+// 1 GB in bytes
+const GB_TO_BYTES = MB_TO_BYTES * 1024;
+
 const lineChartOptions = Object.assign({
     scales: {
         xAxes: [{
@@ -76,7 +81,7 @@ const reqLineChartOptions = Object.assign({
     }
 }, defaultChartOptions);
 
-const dbSizeLineChartOpts = Object.assign({}, {
+const dbSizeLineChartOptsMB = Object.assign({}, {
     scales: {
         xAxes: [{
             ticks: {
@@ -98,9 +103,8 @@ const dbSizeLineChartOpts = Object.assign({}, {
                 maxTicksLimit: 4,
                 suggestedMin: 0,
                 beginAtZero: true,
-                suggestedMax: 100,
                 callback: function (value, index, values) {
-                    return `${value} MB`;
+                    return `${(value / MB_TO_BYTES).toFixed(2)} MB`;
                 }
             },
         }],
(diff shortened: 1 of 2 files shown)


alexsporn (approved):

plugins/dashboard/frontend/src/app/components/Misc.tsx
```diff
@@ -109,12 +113,39 @@ const dbSizeLineChartOpts = Object.assign({}, {
...
+            gridLines: {
+                display: false
+            },
+            ticks: {
+                fontSize: 10,
+                maxTicksLimit: 4,
+                suggestedMin: 0,
+                beginAtZero: true,
+                callback: function (value, index, values) {
+                    return `${(value / GB_TO_BYTES).toFixed(2)} GB`;
```
alexsporn, line 134: Why not check here for the size and decide if MB or GB?

plugins/dashboard/frontend/src/app/components/Misc.tsx
```diff
@@ -109,12 +113,39 @@ const dbSizeLineChartOpts = Object.assign({}, {
...
+                }
+            },
+        }],
+    },
+    tooltips: {
+        callbacks: {
+            label: function (tooltipItem, data) {
+                let label = data.datasets[tooltipItem.datasetIndex].label;
+                return `${label}: ${(tooltipItem.value / GB_TO_BYTES).toFixed(2)} GB`;
```
alexsporn, line 143: Same here

alexsporn (dismissed):

plugins/dashboard/frontend/src/app/components/Misc.tsx
muXxer: @rajivshah3 ?

rajivshah3: Sorry, just saw this. The reason why is because `dbSizeChartOpts` is defined outside of the class, and the only way for it to know the current size of the database is if it has access to `this.props.nodeStore.last_dbsize_metric`, which is an instance variable. 

Earlier, I tried doing it through the callback but it resulted in the graph showing some values (0-1 GB) in MB and others in GB. This is because (as mentioned above) the callback doesn't know about the rest of the data. The workaround for this was to get the current size of the database and show all values as MB or GB, depending on what the current db size is.

muXxer (approved):

PR #621, opened 2020-08-17
Fix transaction root snapshot indexes calculation

muXxer:
This PR fixes using the wrong index for the solid entry points in the OTRSI, YTRSI calculation for URTS of the nodes and BMD of the COO.

It also simplifies the score calculation of URTS.

### DIFF
--- a/plugins/webapi/tips.go
+++ b/plugins/webapi/tips.go
@@ -113,48 +113,15 @@ func getTipInfo(i interface{}, c *gin.Context, _ <-chan struct{}) {
 		return
 	}
 
-	cachedBundle := tangle.GetCachedBundleOrNil(cachedTxMeta.GetMetadata().GetTxHash()) // bundle +1
-	if cachedBundle == nil {
-		e.Error = "unknown bundle"
-		c.JSON(http.StatusBadRequest, e)
+	// if the OTRSI to LSMI delta is over MaxDeltaTxOldestRootSnapshotIndexToLSMI, the tip is semi-lazy and should be promoted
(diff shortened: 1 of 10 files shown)


alexsporn (approved):

plugins/webapi/tips.go
```diff
@@ -113,49 +113,16 @@ func getTipInfo(i interface{}, c *gin.Context, _ <-chan struct{}) {
...
-	cachedBundle := tangle.GetCachedBundleOrNil(cachedTxMeta.GetMetadata().GetTxHash()) // bundle +1
-	if cachedBundle == nil {
-		e.Error = "unknown bundle"
-		c.JSON(http.StatusBadRequest, e)
+	// if the OTRSI to LSMI delta is over MaxDeltaTxOldestRootSnapshotIndexToLSMI, the tip is semi-lazy and should be promoted
+	if (lsmi - ortsi) > milestone.Index(config.NodeConfig.GetInt(config.CfgTipSelMaxDeltaTxOldestRootSnapshotIndexToLSMI)) {
+		c.JSON(http.StatusOK, GetTipInfoReturn{
+			Confirmed:      false,
+			Conflicting:    false,
+			ShouldPromote:  true,
```
GalRogozinski, line 121: If a tip is semi-lazy it can be promoted?
I thought it couldn't because the ORTSI doesn't change

muXxer: The promotion tx will also land in the semi-lazy pool, but the chance slightly increases that it will be picked.

PR #622, opened 2020-08-18
Mainnet & Devnet configs for Chrysalis Pt1

karimodm:
# Description

Here are the _proposed_ config files for DevNet and MainNet for Chrysalis Pt1.
Configuration blocks that do not set specific values relevant for tipselection or coordinator have been selected with sane defaults for end-user usability.

## Rationale for MainNet values

`intervalSeconds`: `15` we want to be fast to confirm quickly and avoid attacks by denying creation of big side-tangles.
`C1`: `8` -> Regular tipselection will reduce the amount of non-lazy tips, as we select a pretty big value here.
`C2`: `13` -> If your device is slow, allow a certain "grace" period for new tips to be considered semi-lazy, having more than a minute (C2 - C1) window buffer to be narrowed by spammers.
`BMD`: `15` -> Way too slow, sorry. The amount of PoW needed to confirm too far past this point is undoable.
`maxHeaviestBranchTipsPerCheckpoint`: `10` -> Let's try to confirm as many tips as possible; assuming a pretty slow and geographically-distributed network with high(ish) lambda we try to collect quite a few tips.
`randomTipsPerCheckpoint`: `3` -> Keep this choice of values quite fluid by allowing a good portion of randomness in the heaviest tipsel.

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,6 +34,7 @@
   "dashboard": {
     "bindAddress": "localhost:8081",
     "theme": "default",
+    "dev": false,
     "basicAuth": {
       "enabled": false,
       "username": "",
@@ -44,6 +45,7 @@
   "snapshots": {
     "loadType": "local",
     "local": {
+      "depth": 50,
       "intervalSynced": 50,
       "intervalUnsynced": 1000,
       "path": "snapshots/mainnet/export.bin",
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -70,6 +72,24 @@
   "spentAddresses": {
     "enabled": true
   },
+  "coordinator": {
+    "address": "UDYXTZBE9GZGPM9SSQV9LTZNDLJIZMPUVVXYXFYVBLIEUHLSEWFTKZZLXYRHHWVQV9MNNX9KZC9D9UZWZ",
+    "securityLevel": 2,
+    "merkleTreeDepth": 24,
```
alexsporn, line 78: @karimodm these values should be the hard-coded defaults in the code instead of in the config.json
We want to try to reduce the amount of values in the default mainnet config and let the node pick the "defaults"

IoTMOD:
Stolen and fixed by @jakubcech with #623

PR #650, opened 2020-09-02
Release v0.5.3 rc1

muXxer:


pkg/testsuite/utils/dot_file.go
```diff
@@ -0,0 +1,62 @@
...
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+
+	stdin.Write([]byte(dotCommand))
+	stdin.Close()
+
+	cmd.Wait()
+
+	if err := exec.Command("xdg-open", outFilePath).Start(); err != nil {
```
alexsporn: this is very linux-only ;-)

muXxer: there are other OS?? :carlos:

plugins/snapshot/local_snapshot.go
```diff
@@ -453,6 +452,7 @@ func createLocalSnapshotWithoutLocking(targetIndex milestone.Index, filePath str
 	if writeToDatabase {
 		// This has to be done before acquiring the SolidEntryPoints Lock, otherwise there is a race condition with "solidifyMilestone"
 		// In "solidifyMilestone" the LedgerLock is acquired, but by traversing the tangle, the SolidEntryPoint Lock is also acquired.
+		// ToDo: we should flush the caches here, just to be sure that all information before this local snapshot we stored in the persistence layer.
```
alexsporn, line 455: future muxxer will  handle that? :carlos:

muXxer: Thats for future hornet team.... obviously thats not easy at all and only a really rare edge case. Just wanted to comment it here so we do know about that.

plugins/snapshot/pruning.go
```diff
@@ -20,7 +20,7 @@ const (
 )
 
 // pruneUnconfirmedTransactions prunes all unconfirmed tx from the database for the given milestone
-func pruneUnconfirmedTransactions(targetIndex milestone.Index) int {
+func pruneUnconfirmedTransactions(targetIndex milestone.Index) (int, int) {
```
alexsporn: give labels to the return values

alexsporn (approved):

pkg/testsuite/utils/dot_file.go
alexsporn: Add a TODO there and each platform dev should add platform support as needed ;-)

plugins/snapshot/local_snapshot.go
GalRogozinski: isn't it just a call to `Tangle.Flush`?

I guess not, because then it would be easy so what am I misunderstanding?

muXxer: We can't call the flush method if other parts of the node still access the objects. Object storage is not yet capable of doing this. I already talked to hans, he wants to change something and implement a "timeheap", and as a side effect this will also be possible hopefully. I added that comment that we don't forget that in the future.

PR #661, opened 2020-09-14
Release v0.5.3

muXxer:
### Added

    - testsuite for easier test case creation
    - ported integration test framework from goshimmer

### Changed

    - added environment variables to /etc/default/hornet for easier node configuration
    - livefeed transactions are now filtered before they are rate limited (dashboard)
    - reduce websocket traffic by adding new message for sync status (dashboard)
    - config files can now be passed with file extension
    - autopeers are now ignored in "maxPeers" check
    - "tag"-Trytes are now padded in "findTransactions" webapi call
    - heartbeats are now used to detect dead peers
    - add write timeout to managed connection

### Fixed

    - connection to peers never closed on error ("waiting" state in dashboard)
    - incorrect alias of peers for incoming connections
    - pruning did not collect all transactions that were referenced by a milestone
    - deadlock if a peer connected during shutdown
    - "node is not synced" glitches in some webapi calls
    - race condition in autopeering shutdown
    - error not handled in SelectSpammerTips
    - pruning routine was exited if there was not enough history at the beginning
    - rpm package upgrade command
    - panic in websocket during connection establishment


docker/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-# golang 1.14.4-buster
+# golang 1.14.6-buster
```
alexsporn, line 1: shouldn't this be 1.15?

integration-tests/tester/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: "3.5"
+
+services:
+  tester:
+    container_name: tester
+    image: golang:1.14.4
```
alexsporn, line 6: 1.15?

alexsporn (approved):

PR #681, opened 2020-09-24
Fix: Do not change default GOMAXPROCS value

### LINKED ISSUE #675: GOMAXPROCS is set to an arbitrary high value
**Describe the bug**

In the database plugin, GOMAXPROCS is set to 128:
https://github.com/gohornet/hornet/blob/13961e391cc0ccfa6b21cd6e7e944d5a0c8a6726/plugins/database/plugin.go#L35 

The GOMAXPROCS variable defines the number of active OS threads that are used by the go scheduler to multipex goroutines into. Setting this to a fixed and very high value regardless of number of cores, OS, architecture, etc. is very dangerous! Especially on hardware with fewer cores, this will lead to many OS threads and much more expensive context switches.

Judging from the comments, it seems like this was introduced as a performance hack for badger DB. Since badger has been replaced, there should be no further motivation for this and leaving GOMAXPROCS unchanged at its default should even increase performance. 
If there happen to be some strong performance gains by modifying GOMAXPROCS, those changes should clearly depend on external hardware factors...

Wollac:
# Description

This PR removes the override of the default `GOMAXPROCS` value. When the corresponding environment variable is not set, leaving `GOMAXPROCS` at its default should lead to reasonable behavior on all systems and platforms.

Fixes #675 

## Type of change

# How Has This Been Tested?

- Starting HORNET and then checking its threads with e.g. `cat /proc/<PROCESS_PID>/status | grep Threads` will now return a value closer to the available CPU threads instead of ≥ 128.
- Note, `GOMAXPROCS` only limits the number of active threads. Threads blocked because of IO operations for examples can lead to a higher number of threads. This is intended behavior.
- I did some benchmark syncs on amd64 VPS, leading to similar sync times and reduced CPU loads.

# Checklist:

- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/plugins/database/plugin.go
+++ b/plugins/database/plugin.go
@@ -1,10 +1,7 @@
 package database
 
 import (
-	"runtime"
 	"time"
-
-	"github.com/spf13/viper"
 
 	"github.com/iotaledger/hive.go/daemon"
 	"github.com/iotaledger/hive.go/logger"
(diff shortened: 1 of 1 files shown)


muXxer (approved):
Thx wollac for testing this.

PR #682, opened 2020-09-24
Fix: Improve tryte distribution of tools/rand-seed

Wollac:
# Description

The trytes of the seed generated with `tools/rand-seed` are not uniformly distributed since 256 is not divisible by 27 (`tryteAlphabetLength`). This slightly reduces the security of the seed.
This PR proposes to use `KerlBytesToTrytes` instead, which works on the entire 48-byte entropy and not on individual trytes. It also assures that the 243rd (last) trit of the hash is always zero, which is a requirement for the key derivation functions implemented in `iotaledger/iota.go`.

## Type of change

# Checklist:

### DIFF
--- a/tools/rand-seed/rand_seed.go
+++ b/tools/rand-seed/rand_seed.go
@@ -5,18 +5,19 @@ import (
 	"fmt"
 
 	"github.com/iotaledger/iota.go/consts"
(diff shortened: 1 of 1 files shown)


muXxer (approved):

muXxer:
Thx Wolfgang!

PR #684, opened 2020-09-28
Use libp2p for networking

luca-moser:
Replaces the networking stack with libp2p.


centercirclesolutions:
Pay no attention to the man behind the curtain

pkg/config/p2p_config.go
```diff
@@ -0,0 +1,22 @@
...
+const (
+	// Defines the bind addresses of this node.
+	CfgP2PBindAddresses = "p2p.bindAddresses"
+	// Defines the path to the peerstore.
+	CfgP2PPeerStorePath = "p2p.peerStore.path"
+	// Defines the high watermark to use within the connection manager.
+	CfgP2PConnMngHighWatermark = "p2p.connectionManager.highWatermark"
+	// Defines the low watermark to use within the connection manager.
+	CfgP2PConnMngLowWatermark = "p2p.connectionManager.lowWatermark"
+	// Defines the  static peers this node should retain a connection to.
```
muXxer: whitespace

pkg/config/p2p_config.go
```diff
@@ -0,0 +1,22 @@
+package config
+
+const (
+	// Defines the bind addresses of this node.
+	CfgP2PBindAddresses = "p2p.bindAddresses"
```
muXxer: Should we name it MultiAddress to signal that it's another format than the rest of the config bindAddresses?

plugins/gossip/sting.go
```diff
@@ -3,99 +3,85 @@ package gossip
...
-			// we can't help this neighbor to become sync, so it's better to drop the connection and free the slots for other peers.
-			log.Infof("dropping autopeered neighbor %s / %s because LSMI (%d) is below our pruning index (%d)", p.Autopeering.Address(), p.Autopeering.ID(), p.LatestHeartbeat.SolidMilestoneIndex, tangle.GetSnapshotInfo().PruningIndex)
-			peering.Manager().Remove(p.ID)
-			return
-		}
-
-		p.Events.HeartbeatUpdated.Trigger(p.LatestHeartbeat)
+		proto.LatestHeartbeat = gossip.ParseHeartbeat(data)
+		proto.HeartbeatReceivedTime = time.Now()
+		proto.Events.HeartbeatUpdated.Trigger(proto.LatestHeartbeat)
```
muXxer, line 68: Do we still have this check of the "below pruning index" somewhere? It was important :)

pkg/p2p/manager.go
```diff
@@ -0,0 +1,671 @@
...
+			return
+
+		case connectPeerMsg := <-m.connectPeerChan:
+			err := m.connectPeer(connectPeerMsg.addrInfo, connectPeerMsg.peerRelation, connectPeerMsg.alias)
+			if err != nil {
+				m.Events.Error.Trigger(fmt.Errorf("error connect to %s (%v): %w", connectPeerMsg.addrInfo.ID.ShortString(), connectPeerMsg.addrInfo.Addrs, err))
+			}
+			if errors.Is(err, ErrPeerInManagerAlready) {
+				m.updateRelation(connectPeerMsg.addrInfo.ID, connectPeerMsg.peerRelation)
+				m.updateAlias(connectPeerMsg.addrInfo.ID, connectPeerMsg.alias)
```
muXxer, line 372: Would that overwrite a known alias if a static peer with an alias gets connected via "autopeering" ?

pkg/p2p/manager.go
```diff
@@ -0,0 +1,671 @@
...
+		return false
+	}
+	return m.host.Network().Connectedness(id) == network.Connected
+}
+
+// calls the given PeerForEachFunc on each Peer.
+func (m *Manager) forEach(f PeerForEachFunc, filter ...PeerRelation) {
+	for _, p := range m.peers {
+		if len(filter) > 0 {
+			if p.Relation != filter[0] {
```
muXxer: maybe better in one if??

pkg/protocol/gossip/service.go
```diff
@@ -0,0 +1,447 @@
...
+	}))
+	// manage libp2p network events
+	s.host.Network().Notify((*netNotifiee)(s))
+	s.eventLoop(shutdownSignal)
+	// de-register libp2p network events
+	s.host.Network().StopNotify((*netNotifiee)(s))
+}
+
+type connectionmsg struct {
+	p    *p2p.Peer
```
muXxer: should be consistent naming between these structs. Sometimes its p, sometimes peer.

`p2p.Peer` should be `peer`
and
`peer.ID` should be `peerID`

pkg/protocol/gossip/service.go
```diff
@@ -0,0 +1,447 @@
...
+		}
+	}
+}
+
+// handles incoming streams and closes them if the given peer's relation should not allow any.
+func (s *Service) handleInboundStream(stream network.Stream) *Protocol {
+	remotePeerID := stream.Conn().RemotePeer()
+	// close if there is already one
+	if _, ongoing := s.streams[remotePeerID]; ongoing {
+		s.closeUnwantedStream(stream)
```
muXxer, line 293: Should the 
`s.Events.InboundStreamCancelled.Trigger(stream)`
be called here as well?

pkg/protocol/gossip/service.go
```diff
@@ -0,0 +1,447 @@
...
+	// "ClosedStream" notifiee handler fire: this is important, because
+	// we want the remote peer to deregister the stream
+	_ = stream.Conn().Close()
+	_ = stream.Reset()
+}
+
+// handles the automatic creation of a protocol instance if the given peer
+// was connected outbound and its peer relation allows it.
+func (s *Service) handleConnected(p *p2p.Peer, conn network.Conn) (*Protocol, error) {
+	// close if there is already one
```
muXxer: wrong comment?

pkg/protocol/gossip/service.go
```diff
@@ -0,0 +1,447 @@
...
+
+// opens up a stream to the given peer.
+func (s *Service) openStream(id peer.ID) (network.Stream, error) {
+	ctx, _ := context.WithTimeout(context.Background(), s.opts.StreamConnectTimeout)
+	stream, err := s.host.NewStream(ctx, id, s.protocol)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create gossip stream to %s: %w", id, err)
+	}
+	// now some special sauce to trigger the remote peer's SetStreamHandler
+	// ¯\_(ツ)_/¯
```
muXxer, line 354: :running_man:

muXxer: Bruder muss los....

muXxer: Aber ganz schnell...

muXxer (changes requested):

muXxer:
Nice work!

pkg/config/p2p_config.go
luca-moser: Sure, lets do `bindMultiAddresses`

pkg/p2p/manager.go
luca-moser: Yeah true it would. I'll add a check to not overwrite the alias in case it isn't empty.

pkg/p2p/manager.go
luca-moser: totally

pkg/protocol/gossip/service.go
luca-moser: Sure, I've also added a reason for the cancellation, so there is a distinction between the two triggers.

pkg/protocol/gossip/service.go
luca-moser: yeah

plugins/gossip/sting.go
luca-moser: I've added it back as a comment within the heartbeat receive handler. We can uncomment it when peer discovery has been implemented.

muXxer (approved):

pkg/protocol/gossip/service.go
luca-moser: Aber jetzt mal ganz schnell dalli dalli

PR #714, opened 2020-11-03
Use dependency injection to setup plugin dependencies

luca-moser:
This PR removes constructors for `Plugin` and `CoreModule`. Node initialization has changed to:
* `Provide`: the plugin/coremodule provides objects to a dependency graph.
* The dependencies the plugin/coremodule wants are initialized.
* `Configure` is called.
* `Run` is called.

Plugins/CoreModule can omit callbacks for the given phases.

i.e.
```go
func init() {
	Plugin = &node.Plugin{
		Name:      "Example",
		DepsFunc:  func(cDeps dependencies) { deps = cDeps },
		Provide:   provide,
		Configure: configure,
		Run:       run,
		Status:    node.Enabled,
	}
}

var (
	Plugin *node.Plugin
	deps   dependencies
)

type dependencies struct {
	dig.In
	Tangle *tangle.Tangle
}

type SomeObject struct{}

// provide objects to the dependency graph which
// this plugin initializes.
func provide(c *dig.Container) {
	if err := c.Provide(func() *SomeObject {
		return &SomeObject{}
	}); err != nil {
		panic(err)
	}
}

func configure() {
	// do something with deps.Tangle
}

func run() {
	// do anything
}
```


core/gossip/core.go
```diff
@@ -35,69 +36,88 @@ const (
...
-func init() {
-	CoreModule = node.NewCoreModule("Gossip", configure, run)
+			Host       host.Host
+			Manager    *p2p.Manager
+			NodeConfig *configuration.Configuration `name:"nodeConfig"`
+		}
+
+		if err := c.Provide(func(deps servicedeps) *gossip.Service {
+			// ToDo: Issa scam! Snapshot info is not yet known here (because snapshot is loaded afterwards)
+			// dafuq?
```
muXxer: DAFAQ!

muXxer (approved):

PR #728, opened 2020-11-09
Actually fix integration tests

luca-moser:
* `common` failed because `IsHealthy` in the info response was false all the time, as it only counted whether `known` peers were connected to deem the node healthy. We're using `unknown` relation between the peers, therefore I've switched it to count instead ongoing gossip streams (which is the more correct metric).
*  `benchmark` failed because the nodes were initialized with proof-of-work disabled via the REST API.

Also fixed:
* Updated iota.go since the node client did not correctly return errors if no response object was expected from a route. This caused things like querying for the `0000...` message even  though the actual submit did not work (this besides the proof-of-work flag not being enabled).
* Made the `PeerErrorCaller` a struct since triggering it without giving it an actual error returns in a nil pointer.
* Bumped the `p2p.Manager` message channels to 10, to prevent in-loop message creation to block it.

### DIFF
--- a/integration-tests/runTests.sh
+++ b/integration-tests/runTests.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-TEST_NAMES='common autopeering benchmark'
+TEST_NAMES='common benchmark'
 
 echo "Build HORNET image"
 docker build -f ../docker/Dockerfile.dev -t hornet:dev ../.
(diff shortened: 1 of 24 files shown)


integration-tests/runTests.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-TEST_NAMES='common autopeering benchmark'
+TEST_NAMES='benchmark'
```
muXxer: Revert me!

muXxer (approved):

integration-tests/runTests.sh
luca-moser: Nope, no autopeering either

Issue #713, opened 2020-11-03
Refactor Tangle and Snapshot plugins


luca-moser:
The Tangle and Snapshot plugins both have a plethora of globals within themselves, such as `Events` structs etc. which are accessed by other plugins.

muXxer:
#729 
#753

PR #729, opened 2020-11-10
Refactor tangle plugin

muXxer:
This PR removes the global states in the tangle plugin.
#713


core/p2p/core.go
```diff
@@ -120,7 +120,6 @@ func provide(c *dig.Container) {
 			)),
 			libp2p.NATPortMap(),
 		)
-		createdHost.ID()
```
muXxer, line 123: Why was this line here @luca-moser ?

pkg/node/node.go
```diff
@@ -12,7 +11,6 @@ import (
 )
 
 type Node struct {
-	wg              *sync.WaitGroup
```
muXxer, line 15: What was this waitGroup used for?

PR #735, opened 2020-11-12
Adds network ID

luca-moser:
* Removes `Version` from `Message`
* Adds `NetworkID` to message which is built from a Blake2b hash of a string of the structure `<network_name><version>`.
* Received/To-Emit Messages must have this `NetworkID` otherwise they are discarded. Peers which sent messages with the wrong network ID are disconnected.
* Adjusted the snapshot tool to take a `<network_name><version>` network ID as a parameter.
* Swapped the snapshot/alphanet/export.bin and integration-tests/assets/snapshot.bin files accordingly.
* Also made the integration tests use the same key pairs as for the network one can setup via the various .sh files.
* If a node consumes a snapshot file, its network ID must correspond to what is configured to the node and panics otherwise.
* If a node starts up and the saved snapshot info in the database does not match the configured network ID, the node panics.
* The gossip-protocol ID string uses now the numeric network ID, so peers with different networks IDs do not peer to each other.


core/protocfg/core.go
```diff
@@ -17,6 +18,8 @@ const (
 	CfgProtocolPublicKeyRanges = "protocol.publicKeyRanges"
 	// the minimum PoW score required by the network
 	CfgProtocolMinPoWScore = "protocol.minPoWScore"
+	// the network ID on which this node operates on.
+	CfgProtocolNetworkIDName = "protocol.networkID"
```
alexsporn, line 22: Why make this configurable? Why not read it out of the snapshot? Makes it more difficult to misconfigure

luca-moser: As discussed in private, we're adding this config option to be analog somewhat to the coordinator address in iota 1.5.

muXxer (approved):

PR #764, opened 2020-11-24
Adds delta snapshot import/export

luca-moser:
* Adds the ability to import/export delta snapshots.
* Adds snapshot integration test testing whether nodes boot up from a full and delta snapshot and resolve to the correct ledger state. The code for the snapshot generation of those two test files can be found under `./tools/intsnap`.
* Separates all the consumer/producer functions from the main import/export functions to increase readability.
* Changed snapshot path strings to `snapshots.fullPath`/`snapshots.deltaPath` respectively.
* The node generates now delta snapshots instead of full snapshots by either writing out milestone diffs from the db if the pruning index allows it, or consumes a previous delta snapshot file and then data from the db. The node must have a full snapshot file while writing out a delta snapshot of which its snapshot index is used as the ledger milestone index within the delta snapshot.
* The node bootstraps now either with:
  * no snapshots, in which case it downloads from `snapshots.downloadURLs` which is now an array of objects of schema: `{"full": "<url>", "delta": "<url>"}`. Note that both files must be downloaded if both URLs are not empty.
  * a full and delta snapshot, in which case it will first read-in the full snapshot, rollback the ledger state to the snapshot index, then read-in the delta snapshot and built up the ledger to the snapshot index of the delta snapshot.
  * a full snapshot, in which case no delta diffs are applied at all
* Added a tool `snapmerge` which merges a full and delta snapshot into a merged full snapshot containing no milestone diffs. It does this by using a temporary Pebble DB with a `utxo.Manager` to replicate the same operations as normal snapshot imports/exports would do. Renders something like this https://ybin.me/p/4d539d2b1de72557#k765AkB6Eo+cGXD5uypBQTQUWjA+WEXPZahAwXHNq1E=
* Added a tool `snapinfo` which outputs the header of a snapshot.


core/snapshot/params.go
```diff
@@ -18,8 +18,10 @@ const (
 	// interval, in milestones, at which snapshot files are created if the ledger is not fully synchronized
 	CfgSnapshotsIntervalUnsynced = "snapshots.intervalUnsynced"
 	// path to the snapshot file
```
muXxer: path to the full snapshot file

pkg/snapshot/download.go
```diff
@@ -55,71 +55,92 @@ func (wc *WriteCounter) PrintProgress() {
...
+	if _, err = io.Copy(out, io.TeeReader(resp.Body, counter)); err != nil {
+		return fmt.Errorf("download failed: %w", err)
 	}
 
-	if err := os.Rename(filepath+".tmp", filepath); err != nil {
-		return err
+	// the progress indicator uses the same line so print a new line once it's finished downloading
+	fmt.Print("\n")
+
+	if err := os.Rename(tempFileName, path); err != nil {
```
muXxer, line 141: Is it ok to rename the file before closing it? :thinking:

pkg/snapshot/snapshot_file.go
```diff
@@ -504,3 +516,176 @@ func readSpent(reader io.Reader) (*Spent, error) {
...
+	pebbleDB, err := pebble.CreateDB(tempDBPath)
+	if err != nil {
+		return nil, err
+	}
+	kvStore := pebble.New(pebbleDB)
+
+	defer func() {
+		// clean up temp db
+		kvStore.Shutdown()
+		_ = pebbleDB.Close()
```
muXxer: kvstore now also contains the Flush and Close methods in the interface :)
<https://github.com/iotaledger/hive.go/blob/master/kvstore/kvstore.go#L144>

pkg/toolset/snapmerge.go
```diff
@@ -0,0 +1,80 @@
...
+	mergeInfo, err := snapshot.MergeSnapshotsFiles(defaultTempDBForMerge, fullPath, deltaPath, targetPath)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("metadata:")
+	printSnapshotHeaderInfo("full", fullPath, mergeInfo.FullSnapshotHeader)
+	printSnapshotHeaderInfo("delta", deltaPath, mergeInfo.DeltaSnapshotHeader)
+	printMergedSnapshotHeaderInfo("merged", targetPath, mergeInfo.MergedSnapshotHeader, mergeInfo)
+	fmt.Printf("successfully created merged full snapshot '%s', took %v\n", args[2], time.Since(start))
```
muXxer: I would add a truncate to Milliseconds to that time.Since

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/koron/go-ssdp v0.0.2 // indirect
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/labstack/echo/v4 v4.1.17
+	github.com/labstack/gommon v0.3.0
```
muXxer: Where is that used?

muXxer (approved):

PR #777, opened 2020-12-02
Adapt alphanet scripts to delta snapshots

muXxer:


luca-moser:
Can we move all this alphanet config and .sh files to another folder? It pollutes the entire root directory with all these files.

muXxer:
sure

PR #803, opened 2020-12-23
Fixes wrong semi-lazy if condition

luca-moser:
This is also what caused the testnet to stagnate as a semi-lazy threshold gets reached and then no node (aka the spammer plugin) will issue messages anymore since the false condition will prevent it from doing so.


muXxer (commented):
lol, sorry for that

PR #832, opened 2021-01-21
docker: use distroless image with glibc as we compile with CGO

karimodm:
# Description

Docker.dev compiles a dinamically-linked go binary using CGO, to fully support `net` functions a glibc needs to be present on the runtime system.

Fixes hostname resolution problems (glibc dependant) with custom docker networks.

## Type of change

# How Has This Been Tested?

Ran in latest Docker Engine (v20)

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -29,7 +29,7 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build \
 ############################
 # using static nonroot image
 # user:group is nonroot:nonroot, uid:gid = 65532:65532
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.dev
```diff
@@ -29,7 +29,7 @@ RUN CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build \
 ############################
 # using static nonroot image
 # user:group is nonroot:nonroot, uid:gid = 65532:65532
```
muXxer: Could you modify the comment, so that this won't happen in the future? Also "static" is wrong now :D

PR #842, opened 2021-01-27
UTXO storage improvements

alexsporn:
- Added description of the UTXO storage keys and values.
- Future-proofed the UTXO storage to support addresses other than Ed25519 without needing db migrations.
- Added the output type to the lookup tables so we can filter.
- Added tests for serialization/deserialization of the UTXO models.
- Added tests for applying a milestone confirmation and rolling back the confirmation.
- Added new iterate options when loading UTXO, i.e. read lock ledger, filter by address, filter by type, max count.
- Added new `type` filter to the outputs rest API endpoints.


pkg/model/utxo/marshalutils_helper.go
```diff
@@ -0,0 +1,68 @@
...
+
+	addr, err := iotago.AddressSelector(uint32(addrType))
+	if err != nil {
+		return nil, err
+	}
+
+	address := addr.(iotago.Address)
+
+	pre := ms.ReadOffset()
+	readBytes, err := address.Deserialize(ms.ReadRemainingBytes(), iotago.DeSeriModePerformValidation)
```
muXxer, line 57: Do we always need to perform validation if we read from our own storage?

pkg/model/utxo/milestone_diff.go
```diff
@@ -0,0 +1,140 @@
...
+	return m.Bytes()
+}
+
+func (ms *MilestoneDiff) kvStorableKey() []byte {
+	return milestoneDiffKeyForIndex(ms.Index)
+}
+
+func (ms *MilestoneDiff) kvStorableValue() []byte {
+
+	m := marshalutil.New(4 + len(ms.Outputs)*34 + len(ms.Spents)*67)
```
muXxer: Isn't there 4 bytes missing for the spent count?

pkg/model/utxo/output.go
```diff
@@ -147,63 +159,101 @@ func NewOutput(messageID *hornet.MessageID, transaction *iotago.Transaction, ind
...
+	if opt.readLockLedger {
+		u.ReadLockLedger()
+		defer u.ReadUnlockLedger()
+	}
 
+	var innerErr error
+	var i int
 	if err := u.utxoStorage.Iterate([]byte{UTXOStoreKeyPrefixOutput}, func(key kvstore.Key, value kvstore.Value) bool {
 
+		if (opt.maxResultCount > 0) && (i+1 > opt.maxResultCount) {
```
muXxer: why not  `(i >= opt.maxResultCount)`? :trollface:

pkg/model/utxo/spent.go
```diff
@@ -123,22 +149,45 @@ func (u *Manager) loadOutputOfSpent(s *Spent) error {
...
+		addrBytes, err := opt.address.Serialize(iotago.DeSeriModeNoValidation)
+		if err != nil {
+			return err
 		}
-		key = byteutils.ConcatBytes(key, address[0][:])
+		key = byteutils.ConcatBytes(key, addrBytes)
 	}
 
+	// Filter by output type
+	if opt.filterOutputType != nil {
```
muXxer: This is wrong.
You can't apply this filter, if the filter for the address was not set before.
Otherwise you iterate over the wrong keys.

You need to move this if into the one above.

pkg/model/utxo/spent.go
```diff
@@ -123,22 +149,45 @@ func (u *Manager) loadOutputOfSpent(s *Spent) error {
...
+		key = byteutils.ConcatBytes(key, []byte{*opt.filterOutputType})
+	}
+
+	var i int
+
 	if err := u.utxoStorage.Iterate(key, func(key kvstore.Key, value kvstore.Value) bool {
 
+		if (opt.maxResultCount > 0) && (i+1 > opt.maxResultCount) {
```
muXxer: why not (i >= opt.maxResultCount)? :trollface:

plugins/restapi/v1/utxo.go
```diff
@@ -210,9 +226,23 @@ func outputsIDsByEd25519Address(c echo.Context) (*addressOutputsResponse, error)
 		return nil, errors.WithMessage(restapi.ErrServiceUnavailable, "node is not synced")
 	}
 
-	addressParam := strings.ToLower(c.Param(ParameterAddress))
 	includeSpent, _ := strconv.ParseBool(strings.ToLower(c.QueryParam("include-spent")))
 
+	var filteredType *iotago.OutputType
```
muXxer, line 231: Should we create a small helper func for this?

pkg/model/utxo/unspent.go
```diff
@@ -35,21 +63,49 @@ func (u *Manager) IsOutputUnspent(outputID *iotago.UTXOInputID) (bool, error) {
...
+		addrBytes, err := opt.address.Serialize(iotago.DeSeriModeNoValidation)
+		if err != nil {
+			return err
 		}
-		key = byteutils.ConcatBytes(key, address[0][:])
+		key = byteutils.ConcatBytes(key, addrBytes)
+	}
+
+	// Filter by type
+	if opt.filterOutputType != nil {
```
muXxer: Same problem like in the spents file.
Output type filter can only be applied if address filter was set.
Maybe return an error if someone tries to filter by output type without address filter set?

pkg/model/utxo/unspent.go
```diff
@@ -35,21 +63,49 @@ func (u *Manager) IsOutputUnspent(outputID *iotago.UTXOInputID) (bool, error) {
...
+		key = byteutils.ConcatBytes(key, []byte{*opt.filterOutputType})
 	}
 
+	var i int
+
 	if err := u.utxoStorage.IterateKeys(key, func(key kvstore.Key) bool {
 
-		outputKey := byteutils.ConcatBytes([]byte{UTXOStoreKeyPrefixOutput}, key[1+iotago.Ed25519AddressBytesLength:])
+		if (opt.maxResultCount > 0) && (i+1 > opt.maxResultCount) {
```
muXxer: `>=`
:runner:

muXxer (changes requested):

pkg/model/utxo/marshalutils_helper.go
alexsporn: It only does some length checks and verifies the type is correct, so nothing expensive

muXxer (approved):

PR #845, opened 2021-01-30
Adds multiple parents, use iota.go/v2, use ZIP compliant Ed25519

muXxer:


luca-moser:
Note that because I've added the v2 module which also includes your fix with the dups, this PR also adds the new receipt models but of course they aren't used yet. And as another consequence, I've let the code use the ZIP compliant Ed25519 package for good measure.

pkg/model/hornet/message_id.go
```diff
@@ -0,0 +1,148 @@
...
+	if len(b) != iotago.MessageIDLength {
+		panic(fmt.Sprintf("unknown messageID length (%d)", len(b)))
+	}
+
+	return MessageID(b)
+}
+
+// MessageIDFromArray creates a MessageID from a byte array.
+func MessageIDFromArray(b iotago.MessageID) MessageID {
+	return MessageID(b[:])
```
Wollac:
```suggestion
	return append(MessageID{}, b[:]...)
```
I would seriously consider copying the actual iotago.MessageID and not just pointing to it.
This has a minimal performance penalty when b is indeed not mutated, but since it can cause nasty bugs (like the loop variable) it might be worth it.

luca-moser (approved):
lgtm

pkg/model/hornet/message_id.go
```diff
@@ -0,0 +1,148 @@
...
+	return results
+}
+
+// ToSliceOfSlices converts the MessageIDs to a slice of byte slices.
+func (m MessageIDs) ToSliceOfSlices() [][]byte {
+	var results [][]byte
+	for _, msgID := range m {
+		// this is necessary, otherwise we create a pointer to the loop variable
+		tmp := msgID
+		results = append(results, tmp)
```
Wollac:
```suggestion
		results = append(results, msgID)
```
Declaring a new variable and thus copying the value of `msgID` has no effect as msgID is a `[]byte`. In go the value of a slice is its _header_ which contains a pointer to the underlying array. (This is different for arrays, where an assignment copies all its values)

pkg/model/hornet/message_id.go
```diff
@@ -0,0 +1,148 @@
...
+	sort.Sort(orderedArray)
+
+	result := make(MessageIDs, uniqueElements)
+	for i, v := range orderedArray {
+		// this is necessary, otherwise we create a pointer to the loop variable
+		tmp := v
+		result[i] = tmp
+	}
+
+	return result
```
Wollac:
```suggestion
	// sort the messages lexicographically
	sorted := make(iotago.LexicalOrderedByteSlices, len(m))
	for i, id := range m {
		sorted[i] = id
	}
	sort.Sort(sorted)

	var result MessageIDs
	var prev []byte
	for i, id := range sorted {
		// only add to the result, if it is different from its predecessor
		if i == 0 || !bytes.Equal(prev, id) {
			result = append(result, id)
		}
		prev = id
	}
	return result
```
This would be an alternative implementation. That could be much cleaner.
It might be a bit slower for huge inputs with many duplicates (as sorting is n*log(n) while map accesses - excluding resizing - are amortized n), while it is faster for inputs with few duplicates. If that is of any consideration, the current solution is fine, although `tmp` is unnecessary.

pkg/model/hornet/message_id.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// MessageIDsFromSliceOfArrays creates slice of MessageIDs from a slice of arrays.
+func MessageIDsFromSliceOfArrays(b iotago.MessageIDs) MessageIDs {
+	result := make(MessageIDs, len(b))
+	for i, msgID := range b {
+		// this is necessary, otherwise we create a pointer to the loop variable
+		tmp := msgID
+		result[i] = tmp[:]
```
Wollac:
```suggestion
		// as msgID is reused between iterations, it must be copied
		result[i] = MessageIDFromArray(msgID)
```
Here, the declaration of `tmp` is indeed crucial, as otherwise all elements of result would point to the same array of the loop variable.
When `MessageIDFromArray` has been changed to a copy, this version might be cleaner though.

pkg/snapshot/snapshot_file.go
```diff
@@ -632,11 +632,11 @@ func MergeSnapshotsFiles(tempDBPath string, fullPath string, deltaPath string, t
 	defer deltaSnapshotFile.Close()
 
 	// build up ledger state to delta snapshot index
-	deltaSnapSEPs := make([]*hornet.MessageID, 0)
+	deltaSnapSEPs := make([]hornet.MessageID, 0)
```
alexsporn: why not `hornet.MessageIDs`

alexsporn (approved):

PR #850, opened 2021-02-02
Adds migration related components

luca-moser:


plugins/restapi/v1/types.go
```diff
@@ -150,6 +156,12 @@ type addressOutputsResponse struct {
 	OutputIDs []string `json:"outputIds"`
 }
 
+// treasuryResponse defines the response of a GET debug treasury REST API call.
```
muXxer: debug?

plugins/restapi/v1/receipts.go
```diff
@@ -0,0 +1,41 @@
...
+	}); err != nil {
+		return nil, errors.WithMessagef(restapi.ErrInternalError, "unable to retrieve receipts: %s", err)
+	}
+
+	return &receiptsResponse{Receipts: receipts}, nil
+}
+
+func receiptsByMigratedAtIndex(c echo.Context) (*receiptsResponse, error) {
+	migratedAtStr := c.Param(ParameterMilestoneIndex)
+	migratedAt, err := strconv.ParseUint(migratedAtStr, 10, 32)
```
muXxer: You could use the function `ParseMilestoneIndexParam` from `v1/milestones.go`.

plugins/receipt/param.go
```diff
@@ -0,0 +1,45 @@
...
+	CfgReceiptsValidatorCoordinatorAddress = "receipts.validator.coordinator.address"
+	// CfgReceiptsValidatorCoordinatorMerkleTreeDepth configures the depth of the Merkle tree of the legacy coordinator.
+	CfgReceiptsValidatorCoordinatorMerkleTreeDepth = "receipts.validator.coordinator.merkleTreeDepth"
+)
+
+var params = &node.PluginParams{
+	Params: map[string]*flag.FlagSet{
+		"nodeConfig": func() *flag.FlagSet {
+			fs := flag.NewFlagSet("", flag.ContinueOnError)
+			fs.String(CfgReceiptsBackupFolder, "./receipts", "path to the receipts backup folder")
```
muXxer: We don't have that `./` in the other paths, like the db for example.
If it is not necessary, maybe we should remove it everywhere to have consistency in the config.

alphanet/migration_bootstrap.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+./create_snapshot_alphanet.sh
+rm -rf alphanetdb
+rm coordinator.state
+export COO_PRV_KEYS=651941eddb3e68cb1f6ef4ef5b04625dcf5c70de1fdc4b1c9eadb2c219c074e0ed3c3f1a319ff4e909cf2771d79fece0ac9bd9fd2ee49ea6c0885c9cb3b1248c,0e324c6ff069f31890d496e9004636fd73d8e8b5bea08ec58a4178ca85462325f6752f5f46a53364e2ee9c4d662d762a81efd51010282a75cd6bd03f28ef349c
+go run -tags "pow_avx" ../main.go -c config_alphanet.json \
```
muXxer: Maybe we could remove the "pow_avx" tags from all scripts :)

pkg/model/migrator/receipt.go
```diff
@@ -0,0 +1,247 @@
...
+// If the receipt has the final flag set to true, then the entire batch of receipts with the same migrated_at index
+// are collected and it is checked whether they migrated all the funds of the given white-flag confirmation.
+func (rs *ReceiptService) Validate(r *iotago.Receipt) error {
+	if !rs.ValidationEnabled {
+		panic("receipt service is not configured to validate receipts")
+	}
+
+	var highestMigratedAtIndex uint32
+	if err := rs.utxoManager.ForEachReceiptTuple(func(rt *utxo.ReceiptTuple) bool {
+		if highestMigratedAtIndex == 0 || rt.Receipt.MigratedAt > highestMigratedAtIndex {
```
muXxer: Why do you need to check for zero?

plugins/migrator/plugin.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func configure() {
+	log = logger.NewLogger(Plugin.Name)
+
+	var msIndex *uint32
+	if *bootstrap {
+		msIndex = startIndex
+	}
+	// TODO: perform sanity check, that the latest migration milestone has MigratedAt lower than the state
```
muXxer: You are still aware of that ToDo? :)

pkg/model/utxo/utxo.go
```diff
@@ -78,6 +78,16 @@ func (u *Manager) PruneMilestoneIndex(msIndex milestone.Index) error {
 		return err
 	}
 
+	// TODO: delete treasury output?
```
muXxer: Is this ToDo still needed?
Do you mean the Spent Treasury Output?

pkg/model/utxo/treasury_output.go
```diff
@@ -0,0 +1,244 @@
...
+	"fmt"
+
+	"github.com/iotaledger/hive.go/kvstore"
+	"github.com/iotaledger/hive.go/marshalutil"
+	iotago "github.com/iotaledger/iota.go/v2"
+)
+
+const (
+	// A prefix which denotes a spent treasury output.
+	TreasuryOutputSpentPrefix = 1
```
muXxer, line 15: We should add a comment that this should not be changed because it uses `WriteBool` to set the byte in the prefix as Spent or not.

pkg/model/migrator/validator.go
```diff
@@ -0,0 +1,215 @@
...
+
+	included, err := m.validateConfirmation(confirmation, milestoneIndex)
+	if err != nil {
+		return nil, fmt.Errorf("invalid confirmation data: %w", &CriticalError{Err: err})
+	}
+
+	migrated := make([]*iota.MigratedFundsEntry, 0, len(included))
+	for i := range included {
+		output := included[i][0]
+		edAddr, _ := address.ParseMigrationAddress(output.Address)
```
muXxer, line 75: Why don't we check for the returned error here?
What if an address can't be parsed?

pkg/model/utxo/milestone_diff.go
```diff
@@ -30,7 +37,7 @@ func (ms *MilestoneDiff) kvStorableKey() []byte {
 
 func (ms *MilestoneDiff) kvStorableValue() []byte {
 
-	m := marshalutil.New(4 + len(ms.Outputs)*34 + 4 + len(ms.Spents)*67)
+	m := marshalutil.New()
```
muXxer, line 40: Why don't we want to pre-allocate the internal byte slice with the minimum size?
Shouldn't this be faster?

pkg/model/utxo/milestone_diff.go
```diff
@@ -43,9 +50,21 @@ func (ms *MilestoneDiff) kvStorableValue() []byte {
 		m.WriteBytes(spent.output.outputID[:])
 	}
 
+	if ms.TreasuryOutput != nil {
+		m.WriteBool(true)
+		m.WriteBytes(ms.TreasuryOutput.MilestoneID[:])
+		// TODO: maybe do another if and check whether any treasury output was spent
```
muXxer: The genesis treasury should also exist in the database IMHO.
But I guess this is not the best place to do that check.
Or would you just panic in case it doesn't exist?

pkg/model/utxo/milestone_diff.go
```diff
@@ -76,6 +95,7 @@ func (ms *MilestoneDiff) kvStorableLoad(utxoManager *Manager, key []byte, value
 
 	spents := make(Spents, spentCount)
 	for i := 0; i < int(spentCount); i++ {
+		// TODO: why are we storing the address with the diff but then don't use it when reading it?
```
muXxer, line 96: I will add it as an open issue to remove that in another PR.

pkg/model/utxo/milestone_diff.go
```diff
@@ -93,6 +113,37 @@ func (ms *MilestoneDiff) kvStorableLoad(utxoManager *Manager, key []byte, value
...
+		return err
+	}
+
+	if hasTreasury {
+		treasuryOutputMilestoneID, err := marshalUtil.ReadBytes(iotago.MilestoneIDLength)
+		if err != nil {
+			return err
+		}
+
+		treasuryOutput, err := utxoManager.readUnspentTreasuryOutputWithoutLocking(treasuryOutputMilestoneID)
```
muXxer, line 126: Couldn't it happen that you load a milestone diff, where the treasure output was already spent?
Maybe try to load the spent output if the unspent failed? :thinking:

pkg/snapshot/snapshot.go
```diff
@@ -772,6 +802,18 @@ func newOutputConsumer(utxoManager *utxo.Manager) OutputConsumerFunc {
 	}
 }
 
+// returns a treasury output consumer which overrides an existing unspent treasury output with the new one.
+func newUnspentTreasuryOutputConsumer(utxoManager *utxo.Manager) UnspentTreasuryOutputConsumerFunc {
+	return func(output *utxo.TreasuryOutput) error {
+		// delete previous
+		existing, err := utxoManager.UnspentTreasuryOutput()
+		if err == nil {
+			_ = utxoManager.DeleteTreasuryOutput(existing)
```
muXxer: Maybe add a comment why we can ignore the error here, because the Treasury exists and there shouldn't be more than one unspent entry.

pkg/snapshot/snapshot.go
```diff
@@ -901,15 +958,21 @@ func (s *Snapshot) LoadSnapshotFromFile(snapshotType Type, networkID uint64, fil
...
 	msDiffConsumer := newMsDiffConsumer(s.utxo)
-	if err := StreamSnapshotDataFrom(lsFile, headerConsumer, sepConsumer, outputConsumer, msDiffConsumer); err != nil {
+	if err := StreamSnapshotDataFrom(lsFile, headerConsumer, sepConsumer, outputConsumer, treasuryOutputConsumer, msDiffConsumer); err != nil {
 		return fmt.Errorf("unable to import %s snapshot file: %w", snapshotNames[snapshotType], err)
 	}
 
 	s.log.Infof("imported %s snapshot file, took %v", snapshotNames[snapshotType], time.Since(ts))
+
+	_ = s.utxo.ForEachTreasuryOutput(func(output *utxo.TreasuryOutput) bool {
+		return true
```
muXxer: Can you add a comment why we call this here without "doing anything"?

muXxer (commented):
Looks very good to me, just some minor questions from my side :+1:

pkg/model/migrator/receipt.go
luca-moser: True, makes no sense

pkg/model/utxo/milestone_diff.go
luca-moser: There shouldn't be any state in which there isn't a treasury in the database as any full snapshot must contain a treasury output. So this TODO is misleading because the genesis treasury would be in the genesis snapshot without a corresponding milestone diff.

pkg/model/utxo/milestone_diff.go
luca-moser: Very good catch! This would have surely ramified itself into a panic if a snapshot would have been produced of which ms diffs contained would have spent treasury outputs.

pkg/snapshot/snapshot.go
luca-moser: Weird, I must have forgotten to remove this code.

pkg/model/migrator/validator.go
luca-moser: I've added a comment but essentially because we're already verifying beforehand that all the bundles are indeed migration bundles and therefore the tail tx is the output tx with the valid migration address.

pkg/model/utxo/treasury_output.go
luca-moser: Added a comment

pkg/model/utxo/milestone_diff.go
luca-moser: Because it doesn't matter speed wise

pkg/model/utxo/utxo.go
luca-moser: Added code to delete the spent treasury output of the corresponding milestone.

plugins/migrator/plugin.go
luca-moser: Yeah it is a todo outstanding for Wolfgang.

pkg/snapshot/snapshot.go
luca-moser: Added a comment

plugins/receipt/param.go
luca-moser: Changed to `receipts`

core/tangle/core.go
```diff
@@ -4,6 +4,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/gohornet/hornet/pkg/model/migrator"
```
alexsporn: sort imports

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/gorilla/websocket v1.4.2
 	github.com/hashicorp/go-version v1.2.1 // indirect
 	github.com/iotaledger/hive.go v0.0.0-20210209220536-c4c55778aaf7
+	github.com/iotaledger/iota.go v1.0.0-beta.15.0.20210212090247-51c40bcebea7
```
alexsporn, line 37: :wen: not beta 😂

pkg/model/coordinator/coordinator.go
```diff
@@ -6,6 +6,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/gohornet/hornet/pkg/model/migrator"
```
alexsporn: sort imports

pkg/model/migrator/receipt.go
```diff
@@ -0,0 +1,247 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/gohornet/hornet/pkg/model/utxo"
```
alexsporn, line 13: split hornet imports into separate group

pkg/model/migrator/receipt.go
```diff
@@ -0,0 +1,247 @@
...
+	ValidationEnabled bool
+	// Whether the service should ignore soft errors.
+	IgnoreSoftErrors bool
+	backupFolder     string
+	validator        *Validator
+	utxoManager      *utxo.Manager
+}
+
+// NewReceiptService creates a new ReceiptService.
+func NewReceiptService(v *Validator, utxoManager *utxo.Manager, validationEnabled bool, backupEnabled bool, ignoreSoftErrors bool, backupFolder string) *ReceiptService {
```
alexsporn: why just `v`?

pkg/model/migrator/service.go
```diff
@@ -0,0 +1,248 @@
+package migrator
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/gohornet/hornet/pkg/utils"
```
alexsporn, line 10: separae hornet imports

pkg/model/migrator/receipt.go
```diff
@@ -0,0 +1,247 @@
...
+		panic("receipt service is not configured to backup receipts")
+	}
+
+	receiptFileName := path.Join(rs.backupFolder, fmt.Sprintf(receiptFilePattern, r.Receipt.MigratedAt, r.MilestoneIndex))
+	receiptJSON, err := r.Receipt.MarshalJSON()
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(receiptFileName, receiptJSON, os.ModePerm); err != nil {
+		return fmt.Errorf("unable to persist receipt onto disk: %w", &SoftError{Err: err})
```
alexsporn: why is not persisting the backup a SoftError?

luca-moser: I believe you're right that the argument can be made, that if someone enabled backups of receipts, then they probably want the node to go hi-wire in case they can't be persisted. Going to change it to a crit err. (It doesn't actually change the error logic handler for this as we always panic up on Backup() if an error is returned).

pkg/model/migrator/service.go
luca-moser: sorted

pkg/model/migrator/receipt.go
luca-moser: renamed to `validator`

pkg/model/migrator/receipt.go
luca-moser: split

pkg/model/coordinator/coordinator.go
luca-moser: sorted

core/tangle/core.go
luca-moser: sorted

go.mod
luca-moser: production ready by EOY 2018

pkg/model/migrator/validator.go
```diff
@@ -0,0 +1,224 @@
+package migrator
+
+import (
+	"bytes"
+	"crypto"
+	"encoding"
+	"fmt"
+
+	"github.com/gohornet/hornet/pkg/whiteflag"
```
alexsporn, line 12: split and sort imports ;-)

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,188 @@
+package utxo
+
+import (
+	"encoding/binary"
+
+	"github.com/gohornet/hornet/pkg/model/hornet"
+	"github.com/gohornet/hornet/pkg/model/milestone"
+	"github.com/iotaledger/hive.go/kvstore"
```
alexsporn, line 9: same procedure here :carlos:

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+
+	rt.Receipt = r
+	rt.MilestoneIndex = milestone.Index(msIndex)
+
+	return nil
+}
+
+// StoreReceipt stores the given receipt tuple.
```
alexsporn: StoreReceipt and DeleteReceipt can be removed. They are unused and are not inside the batched mutations, so should be avoided.

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,188 @@
...
+	return mutations.Delete(rt.kvStorableKey())
+}
+
+// ReceiptTupleConsumer is a function that consumes a receipt tuple.
+type ReceiptTupleConsumer func(rt *ReceiptTuple) bool
+
+// ForEachReceiptTuple iterates over all stored receipt tuples.
+func (u *Manager) ForEachReceiptTuple(consumer ReceiptTupleConsumer, options ...UTXOIterateOption) error {
+	opt := iterateOptions(options)
+
```
alexsporn, line 99: do we want to add this here to be consistent?
```
	if opt.readLockLedger {
		u.ReadLockLedger()
		defer u.ReadUnlockLedger()
	}
```

Same with `ForEachMigratedAtReceiptTuple`

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,188 @@
...
+		return consumer(rt)
+	}); err != nil {
+		return err
+	}
+
+	return innerErr
+}
+
+// ForEachMigratedAtReceiptTuple iterates over all stored receipt tuples for a given migrated at index.
+func (u *Manager) ForEachMigratedAtReceiptTuple(migratedAtIndex uint32, consumer ReceiptTupleConsumer, options ...UTXOIterateOption) error {
```
alexsporn: Why not `ForEachReceiptTupleMigratedAt`?

alexsporn: migratedAtIndex uint32 -> use milestone.Index

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,188 @@
...
+
+// TreasuryTransactionOutputAmount returns the amount of the treasury output within the treasury transaction.
+func TreasuryTransactionOutputAmount(tx *iotago.TreasuryTransaction) uint64 {
+	return tx.Output.(*iotago.TreasuryOutput).Amount
+}
+
+// ReceiptToTreasuryMutation converts a receipt to a treasury mutation tuple.
+func ReceiptToTreasuryMutation(r *iotago.Receipt, unspentTreasuryOutput *TreasuryOutput, newMsId *iotago.MilestoneID) (*TreasuryMutationTuple, error) {
+	newOutput := &TreasuryOutput{
+		Amount: r.Transaction.(*iotago.TreasuryTransaction).Output.(*iotago.TreasuryOutput).Amount,
```
alexsporn, line 190: why not use `TreasuryTransactionOutputAmount` instead? Else that func is unused, so remove it.

pkg/model/utxo/treasury_output.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (u *Manager) readSpentTreasuryOutputWithoutLocking(msHash []byte) (*TreasuryOutput, error) {
+	key := append([]byte{UTXOStoreKeyPrefixTreasuryOutput, TreasuryOutputSpentPrefix}, msHash...)
+	val, err := u.utxoStorage.Get(key)
+	if err != nil {
+		return nil, err
+	}
+	to := &TreasuryOutput{}
+	if err := to.kvStorableLoad(nil, key, val); err != nil {
```
alexsporn: I would still pass `u` here. If for some reason we use it in the future we will not understand why it is not set.

pkg/model/utxo/treasury_output.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (u *Manager) readUnspentTreasuryOutputWithoutLocking(msHash []byte) (*TreasuryOutput, error) {
+	key := append([]byte{UTXOStoreKeyPrefixTreasuryOutput, TreasuryOutputUnspentPrefix}, msHash...)
+	val, err := u.utxoStorage.Get(key)
+	if err != nil {
+		return nil, err
+	}
+	to := &TreasuryOutput{}
+	if err := to.kvStorableLoad(nil, key, val); err != nil {
```
alexsporn: Same here, pass `u`

pkg/model/utxo/treasury_output.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil, err
+	}
+	return to, nil
+}
+
+// StoreTreasuryOutput adds the given treasury output to the database.
+func (u *Manager) StoreTreasuryOutput(to *TreasuryOutput) error {
```
alexsporn: I would remove calls not using a batched mutation.
This is called from `snapshot.newUnspentTreasuryOutputConsumer` and it calls delete and store. We should do that in a batch.

pkg/model/utxo/treasury_output.go
```diff
@@ -0,0 +1,246 @@
...
+	return u.utxoStorage.Set(to.kvStorableKey(), to.kvStorableValue())
+}
+
+// DeleteTreasuryOutput deletes the given treasury output from the database.
+func (u *Manager) DeleteTreasuryOutput(to *TreasuryOutput) error {
+	return u.utxoStorage.Delete(to.kvStorableKey())
+}
+
+// Returns the unspent treasury output.
+func (u *Manager) UnspentTreasuryOutput() (*TreasuryOutput, error) {
```
alexsporn: `WithoutLocking`?

pkg/restapi/restapi.go
```diff
@@ -20,6 +20,9 @@ var (
 
 	// ErrServiceUnavailable defines the service unavailable error.
 	ErrServiceUnavailable = errors.New("service unavailable")
+
+	// ErrServiceUnavailable defines the service not implemented error.
```
alexsporn: Copy-paste error

pkg/tangle/tangle.go
```diff
@@ -5,6 +5,7 @@ import (
 	"runtime"
 	"sync"
 
+	"github.com/gohornet/hornet/pkg/model/migrator"
```
alexsporn: move 7 lines down

pkg/whiteflag/confirmation.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/gohornet/hornet/pkg/model/milestone"
 	"github.com/gohornet/hornet/pkg/model/storage"
 	"github.com/gohornet/hornet/pkg/model/utxo"
+	iotago "github.com/iotaledger/iota.go/v2"
```
alexsporn, line 14: :weirdsmileguy:

plugins/coordinator/plugin.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/gohornet/hornet/pkg/model/migrator"
+	"github.com/gohornet/hornet/pkg/model/utxo"
```
alexsporn: ⬇️

plugins/migrator/params.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/gohornet/hornet/pkg/node"
+	flag "github.com/spf13/pflag"
+)
+
+const (
+	// CfgMigratorStateFilePath configures the path to the state file of the migrator.
+	CfgMigratorStateFilePath = "migrator.stateFilePath"
+	// CfgMigratorQueryCooldownPeriod configures the cooldown period for the service to ask for new data
+	// from the legacy node in case the migrator encounters an error.
+	CfgMigratorQueryCooldownPeriod = "migrator.queryCooldownPeriod"
```
alexsporn, line 15: In seconds? what is the unit here?

plugins/migrator/plugin.go
```diff
@@ -0,0 +1,108 @@
+package migrator
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/gohornet/hornet/core/gracefulshutdown"
```
alexsporn: https://tenor.com/7HU3.gif

plugins/prometheus/plugin.go
```diff
@@ -8,6 +8,8 @@ import (
 	"time"
 
 	"github.com/gin-gonic/gin"
+	"github.com/gohornet/hornet/pkg/model/migrator"
+	"github.com/gohornet/hornet/pkg/tangle"
```
alexsporn: https://tenor.com/6a6x.gif

plugins/receipt/param.go
```diff
@@ -0,0 +1,45 @@
...
+
+var params = &node.PluginParams{
+	Params: map[string]*flag.FlagSet{
+		"nodeConfig": func() *flag.FlagSet {
+			fs := flag.NewFlagSet("", flag.ContinueOnError)
+			fs.String(CfgReceiptsBackupFolder, "receipts", "path to the receipts backup folder")
+			fs.Bool(CfgReceiptsBackupEnabled, false, "whether to backup receipts in the backup folder")
+			fs.Bool(CfgReceiptsValidatorIgnoreSoftErrors, false, "whether to ignore soft errors and not panic if one is encountered")
+			fs.Bool(CfgReceiptsValidatorValidate, false, "whether to validate receipts")
+			fs.String(CfgReceiptsValidatorAPIAddress, "http://localhost:14265", "address of the legacy node API")
```
alexsporn: This is the default port we also use for the node. Should we maybe use different defaults?

plugins/receipt/param.go
```diff
@@ -0,0 +1,45 @@
...
+	// CfgReceiptsBackupFolder configures the path to the receipts backup folder.
+	CfgReceiptsBackupFolder = "receipts.backup.folder"
+	// CfgReceiptsValidatorValidate configures whether to validate receipts.
+	CfgReceiptsValidatorValidate = "receipts.validator.validate"
+	// CfgReceiptsValidatorIgnoreSoftErrors configures the node to not panic if a soft error is encountered.
+	CfgReceiptsValidatorIgnoreSoftErrors = "receipts.validator.ignoreSoftErrors"
+	// CfgReceiptsValidatorAPIAddress configures the address of the legacy node API to query for white-flag confirmation data.
+	CfgReceiptsValidatorAPIAddress = "receipts.validator.api.address"
+	// CfgReceiptsValidatorAPITimeout configures the timeout of API calls.
+	CfgReceiptsValidatorAPITimeout = "receipts.validator.api.timeout"
```
alexsporn, line 22: Document the unit of the timeout

plugins/receipt/plugin.go
```diff
@@ -0,0 +1,99 @@
+package receipt
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gohornet/hornet/pkg/model/storage"
```
alexsporn: https://tenor.com/wooJ.gif

alexsporn (changes requested):
See comments above

plugins/migrator/params.go
muXxer: Ignore Alex comment here, because I have another PR opened, which removes all "Seconds" etc from the config settings, and uses the `Duration` notation you used here.

plugins/receipt/param.go
muXxer: Don't do it :)

muXxer:
Please Move the SoftError and Critical error after merging Wollacs PRs to `pkg/common/errors.go`, so we can reuse it in the coordinator for the soft shutdown.

pkg/model/utxo/treasury_output.go
luca-moser: Added `StoreUnspentTreasuryOutput` which uses a batched mutations to delete an existing unspent treasury output and store a new one.

pkg/restapi/restapi.go
luca-moser: good catch

pkg/model/utxo/treasury_output.go
luca-moser: Added the suffix

plugins/receipt/param.go
luca-moser: I'll change it to 14266 though there isn't really any sensible default because either:
- it will be an actual domain
- we don't know what people will configure

pkg/model/utxo/receipt.go
```diff
@@ -0,0 +1,194 @@
...
+
+// ForEachReceiptTupleMigratedAt iterates over all stored receipt tuples for a given migrated at index.
+func (u *Manager) ForEachReceiptTupleMigratedAt(migratedAtIndex milestone.Index, consumer ReceiptTupleConsumer, options ...UTXOIterateOption) error {
+	opt := iterateOptions(options)
+
+	prefix := make([]byte, 5)
+	prefix[0] = UTXOStoreKeyPrefixReceipts
+	binary.LittleEndian.PutUint32(prefix[1:], uint32(migratedAtIndex))
+
+	var innerErr error
```
alexsporn, line 142: Add this one here too:

```
if opt.readLockLedger {
        u.ReadLockLedger()
        defer u.ReadUnlockLedger()
    }
```

alexsporn (approved):

PR #856, opened 2021-02-05
Chore/code cleanups

muXxer:
This PR reduces memory allocations for slices by initializing them with correct capacity if possible.
It also fixes the usage of the wrong address type (was hardcoded).

### DIFF
--- a/plugins/restapi/v1/utxo.go
+++ b/plugins/restapi/v1/utxo.go
@@ -92,7 +92,7 @@ func ed25519Balance(address *iotago.Ed25519Address) (*addressBalanceResponse, er
 	}
 
 	return &addressBalanceResponse{
-		AddressType: iotago.AddressEd25519,
+		AddressType: address.Type(),
 		Address:     address.String(),
 		Balance:     balance,
 	}, nil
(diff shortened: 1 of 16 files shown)


plugins/restapi/v1/utxo.go
```diff
@@ -174,13 +175,14 @@ func outputsResponse(address iotago.Address, includeSpent bool, filterType *iota
 			return nil, errors.WithMessagef(restapi.ErrInternalError, "reading spent outputs failed: %s, error: %s", address, err)
 		}
 
-		for _, spent := range spents {
-			outputIDs = append(outputIDs, spent.OutputID().ToHex())
+		outputIDs = append(outputIDs, make([]string, len(spents))...)
+		for i, spent := range spents {
+			outputIDs[unspentOutputsLength+i] = spent.OutputID().ToHex()
```
alexsporn: why not create a new slice first, fill it up with the spents and THEN append it?

muXxer: we can also do that, I took this one from the golang page <https://github.com/golang/go/wiki/SliceTricks#extend>
Should I change it?

alexsporn: I would change it. It is the same if you expand with an "empty" slice or a "filled" slice. And the `outputIDs[unspentOutputsLength+i]` makes it error prone I think.

alexsporn (approved):

PR #867, opened 2021-02-07
Add new debug plugin for debug API routes

muXxer:
This PR moves all REST routes from plugins to `/api/plugins/<name>`.
It also creates a new plugin for all debug API endpoints.

### DIFF
--- a/integration-tests/tester/framework/config.go
+++ b/integration-tests/tester/framework/config.go
@@ -229,12 +229,7 @@ func DefaultRestAPIConfig() RestAPIConfig {
 			"/api/v1/addresses/ed25519/:address/outputs",
 			"/api/v1/peers/:peerID",
 			"/api/v1/peers",
-			"/api/v1/debug/outputs",
-			"/api/v1/debug/outputs/unspent",
-			"/api/v1/debug/outputs/spent",
-			"/api/v1/debug/ms-diff/:milestoneIndex",
-			"/api/v1/debug/requests",
-			"/api/v1/debug/message-cones/:messageID",
+			"/api/plugins/*",
 		},
 		EnableProofOfWork: true,
 	}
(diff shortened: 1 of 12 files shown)


integration-tests/tester/framework/config.go
```diff
@@ -229,12 +229,12 @@ func DefaultRestAPIConfig() RestAPIConfig {
 			"/api/v1/addresses/ed25519/:address/outputs",
 			"/api/v1/peers/:peerID",
 			"/api/v1/peers",
-			"/api/v1/debug/outputs",
-			"/api/v1/debug/outputs/unspent",
-			"/api/v1/debug/outputs/spent",
-			"/api/v1/debug/ms-diff/:milestoneIndex",
-			"/api/v1/debug/requests",
-			"/api/v1/debug/message-cones/:messageID",
+			"/api/plugins/debug/outputs",
```
alexsporn: Maybe add a wildcard here for all plugins?

plugins/dashboard/routes.go
```diff
@@ -101,7 +101,7 @@ func passThroughAPIRoute(e echo.Context) error {
 func passThroughSpammerRoute(e echo.Context) error {
 	apiBindAddr := deps.NodeConfig.String(restapi.CfgRestAPIBindAddress)
 
-	data, err := readDataFromURL("http://" + apiBindAddr + "/spammer?" + e.Request().URL.RawQuery)
+	data, err := readDataFromURL("http://" + apiBindAddr + "/api/plugins/spammer?" + e.Request().URL.RawQuery)
```
alexsporn, line 104: This probably can be removed if the dashboard adapts to using the api/plugins route

muXxer: Why was it there in the first place? Maybe I will leave it that way and you will change it anyway with the JWT stuff, right?

alexsporn (approved):

PR #882, opened 2021-02-18
Updates signing, removes WOTS, changes indexation index

luca-moser:


pkg/spammer/spammer.go
```diff
@@ -70,7 +70,7 @@ func (s *Spammer) DoSpam(shutdownSignal <-chan struct{}) (time.Duration, time.Du
 	iotaMsg := &iotago.Message{
 		NetworkID: s.networkID,
 		Parents:   tips.ToSliceOfArrays(),
-		Payload:   &iotago.Indexation{Index: indexation, Data: []byte(messageString)},
+		Payload:   &iotago.Indexation{Index: []byte(indexation), Data: []byte(messageString)},
```
muXxer: Should we encode the index as UTF8 here, because we are "the client lib" here somehow :)

pkg/model/storage/indexation.go
```diff
@@ -3,35 +3,36 @@ package storage
...
+// PadIndexationIndex returns a padded indexation index.
+func PadIndexationIndex(index []byte) []byte {
+	return append(index, make([]byte, IndexationIndexLength-len(index))...)
+}
+
 type Indexation struct {
 	objectstorage.StorableObjectFlags
-	indexationHash hornet.MessageID
-	messageID      hornet.MessageID
+	Index     []byte
```
muXxer: why is `Index` public now?

pkg/model/storage/indexation.go
```diff
@@ -41,11 +42,11 @@ func (i *Indexation) GetMessageID() hornet.MessageID {
 // ObjectStorage interface
 
 func (i *Indexation) Update(_ objectstorage.StorableObject) {
-	panic(fmt.Sprintf("Indexation should never be updated: %v, MessageID: %v", i.indexationHash.ToHex(), i.messageID.ToHex()))
+	panic(fmt.Sprintf("Indexation should never be updated: %v, MessageID: %v", i.Index, i.messageID.ToHex()))
```
muXxer: Should we print the bytes array here or the "string()"? :shrug:

pkg/spammer/spammer.go
luca-moser: wdym encode as utf-8? indexation is already a string - which unless users do something weird - it should be a valid utf-8 string

pkg/model/storage/indexation.go
luca-moser: gonna print hex

integration-tests/runTests.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-TEST_NAMES='common value benchmark snapshot'
+TEST_NAMES='value'
```
muXxer: not sure if you should push that, next time I maybe miss that my local tests are not running :D

pkg/testsuite/utils/dot_file.go
```diff
@@ -39,7 +39,7 @@ func ShortenedIndex(cachedMessage *storage.CachedMessage) string {
 		return fmt.Sprintf("%s (%d)", index, conflict)
 	}
 
-	return index
+	return string(index)
```
muXxer: can you encode it to hex after the line 32
`	index := indexation.Index`

plugins/restapi/v1/messages.go
```diff
@@ -168,8 +169,13 @@ func messageIDsByIndex(c echo.Context) (*messageIDsByIndexResponse, error) {
 		return nil, errors.WithMessage(restapi.ErrInvalidParameter, "query parameter index empty")
 	}
 
+	indexBytes, err := hex.DecodeString(index)
```
muXxer, line 173: If some pleb passes `index` longer than 64 bytes, then the `GetIndexMessageIDs` will call `PadIndexationIndex`, which will call a `make` with a negative count...
```go
// PadIndexationIndex returns a padded indexation index.
func (index []byte) []byte {
	return append(index, make([]byte, IndexationIndexLength-len(index))...)
}``

pkg/spammer/spammer.go
muXxer: Can you limit the size of `index` and `indexSemiLazy` in the plugin where it is loaded from the config to max 64 bytes, or maybe better here in the package?

muXxer (changes requested):

muXxer (approved):

Issue #787, opened 2020-12-06
Add JWT auth and create separated login area in dashboard


muXxer:

muXxer:
Added with #883

PR #883, opened 2021-02-18
New Public/Private Dashboard

alexsporn:
This PR adds the new dashboard with login support.
Basic auth was replaced with a login mask and JWT authorization.
This PR also secures the exposed API endpoints and allows read-only access to the dashboard API if not using a JWT token.
The default configuration is a zeroed out hashed password and salt to avoid having default passwords deployed everywhere.

The included dashboard version includes all the updated indexation, address type and hashing changes.


alexsporn:
This closes #787

plugins/dashboard/routes.go
```diff
@@ -23,90 +21,141 @@ const (
...
+				return false
+			}
 		}
-		return e.Blob(http.StatusOK, contentType, data)
+
+		return true
 	}
-	staticBlob, err := appBox.Find(e.Request().URL.Path)
+
+	apiBindAddr := deps.NodeConfig.String(restapi.CfgRestAPIBindAddress)
```
muXxer, line 121: maybe split the port and use localhost internally?

plugins/dashboard/jwt.go
```diff
@@ -0,0 +1,154 @@
...
+	return nil
+}
+
+func (c *jwtClaims) compare(field string, expected string) bool {
+	if field == "" {
+		return false
+	}
+	if subtle.ConstantTimeCompare([]byte(field), []byte(expected)) != 0 {
+		return true
+	} else {
```
muXxer: you can remove the `else`

plugins/dashboard/routes.go
```diff
@@ -23,90 +21,141 @@ const (
...
+	notFoundMiddleware := func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) (err error) {
+			return ErrNotFound
+		}
+	}
+
+	return []echo.MiddlewareFunc{
+		jwtAuth.Middleware(jwtAuthSkipper),
+		middleware.ProxyWithConfig(config),
+		notFoundMiddleware,
```
muXxer, line 162: Do we need this middleware?

plugins/dashboard/routes.go
```diff
@@ -139,176 +188,53 @@ func enforceMaxOneDotPerURL(next echo.HandlerFunc) echo.HandlerFunc {
...
-		}
-	}()
 
-	// this function sends the initial values for some topics
-	sendInitValue := func(client *websockethub.Client, initValuesSent map[byte]struct{}, topic byte) {
-		if _, sent := initValuesSent[topic]; sent {
-			return
+	if len(request.JWT) > 0 {
+		// Verify JWT is still valid
+		if !jwtAuth.VerifyJWT(request.JWT) {
```
muXxer, line 213: Can we do an early return if the auth is still correct, or do we need to issue a new one?

```go
	return c.JSON(http.StatusOK, map[string]string{
		"jwt": request.JWT,
	})``

PR #907, opened 2021-02-25
GetWhiteFlagConfirmation Mock

Wollac:
This PR adds a simple mock to be used for integration tests. It uses the provided `config.json` to generate migrations and milestones then returned by `getWhiteFlagConfirmation` API calls.


integration-tests/mock/pkg/http/whiteflag/init.go
```diff
@@ -0,0 +1,245 @@
...
+				bndl[i], bndl[j] = bndl[j], bndl[i]
+			}
+			bundles = append(bundles, bndl)
+		}
+		includedBundles[msIndex] = bundles
+	}
+	return includedBundles, nil
+}
+
+func generateMigraionAddress(bytes []byte) (trinary.Hash, error) {
```
luca-moser: typo

integration-tests/mock/pkg/http/whiteflag/init.go
```diff
@@ -0,0 +1,245 @@
...
+				Inputs: inputs,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to create bundle: %w", err)
+			}
+			// PrepareTransfers returns the transaction in the reversed order, so we must reverse again
+			for i, j := 0, len(bndl)-1; i < j; i, j = i+1, j-1 {
+				bndl[i], bndl[j] = bndl[j], bndl[i]
+			}
+			bundles = append(bundles, bndl)
```
luca-moser: I've noticed that we don't PoW the "migration bundles", do you think it makes sense to still do that?

luca-moser (changes requested):
I think we should merge this into `migration` because it is more part of C2 than the legacy stuff (don't know how we could use it otherwise in the C2 codebase). Would also be nice to have a Dockerfile so the mock server can be integrated into the integration framework.

integration-tests/mock/pkg/http/whiteflag/init.go
Wollac: I guess, it does not hurt to PoW them, if that is not relevant for the testcase one can set the MWM to 0 anyways.

Wollac:
Moved to own repo

PR #920, opened 2021-03-02
Feat/coo quorum

muXxer:
This PR adds an optional quorum to the coordinator.
With the quorum, other nodes can be asked for their perception of the ledger state for the tips the coordinator has chosen, before a milestone is issued.

If the quorum has at least one node that has another perception, the coordinator is shut down for security reason.
If none of the nodes of a quorum groups answered in time, the milestone is not issued.

Statistics about the quorum are issued via prometheus.

### DIFF
--- a/pkg/model/coordinator/quorum.go
+++ b/pkg/model/coordinator/quorum.go
@@ -1,0 +1,235 @@
+package coordinator
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/gohornet/hornet/pkg/common"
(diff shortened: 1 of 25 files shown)


pkg/model/coordinator/quorum.go
```diff
@@ -0,0 +1,235 @@
...
+
+	quorumStatsLock syncutils.RWMutex
+}
+
+// newQuorum creates a new quorum, which is used to check the correct ledger state of the coordinator.
+// If no groups are given, nil is returned.
+func newQuorum(quorumGroups map[string][]*QuorumClientConfig, timeout time.Duration) *quorum {
+	if len(quorumGroups) == 0 {
+		// coordinator quorum is disabled
+		return nil
```
alexsporn: this should trigger an error imho

alexsporn (approved):

PR #936, opened 2021-03-06
Add gossip on solidification

muXxer:
This PR adds gossip on solidification and punishes neighbors for invalid gossip.


pkg/model/storage/storage.go
```diff
@@ -75,16 +77,18 @@ type Storage struct {
...
 	utxoManager := utxo.New(store)
 
 	s := &Storage{
-		databaseDir: databaseDirectory,
-		store:       store,
-		utxoManager: utxoManager,
+		databaseDir:   databaseDirectory,
+		store:         store,
+		utxoManager:   utxoManager,
+		belowMaxDepth: milestone.Index(belowMaxDepth),
```
alexsporn, line 88: why is this a milestone.Index?

muXxer: all the calculations we do with belowMaxDepth are based on milestones. So I don't need to convert later on.

alexsporn (approved):

PR #941, opened 2021-03-09
Add memcaches for messages and metadata

muXxer:
This PR simplifies the logic for sharing cached objects over different functions/scopes.


luca-moser (approved):
Nice improvement, makes the code more tight

PR #952, opened 2021-03-12
WIP: Add new caching mechanism based on tangle structure 

muXxer:
This PR adds a new caching mechanism, that holds references on cached objects, based on the structure of the tangle.
This should improve overall performance of the node, since we don't rely on the cache times anymore, but on the latest cone of the tangle.


muXxer:
Closed because we will solve this differently.

PR #956, opened 2021-03-12
Makes migrator receipt entries count configurable

luca-moser:
Default set to 110.

### DIFF
--- a/plugins/migrator/plugin.go
+++ b/plugins/migrator/plugin.go
@@ -6,6 +6,8 @@ import (
 	flag "github.com/spf13/pflag"
 	"go.uber.org/dig"
 
+	iotago "github.com/iotaledger/iota.go/v2"
+
 	"github.com/iotaledger/hive.go/timeutil"
 
 	"github.com/gohornet/hornet/core/gracefulshutdown"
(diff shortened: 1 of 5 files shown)


plugins/migrator/plugin.go
```diff
@@ -67,7 +69,20 @@ func provide(c *dig.Container) {
 		Validator  *migrator.Validator
 	}
 	if err := c.Provide(func(deps serviceDependencies) (*migrator.MigratorService, error) {
-		return migrator.NewService(deps.Validator, deps.NodeConfig.String(CfgMigratorStateFilePath)), nil
+
+		maxReceiptEntries := deps.NodeConfig.Int(CfgMigratorReceiptMaxEntries)
+		switch {
+		case maxReceiptEntries > iotago.MaxMigratedFundsEntryCount:
+			panic(fmt.Sprintf("%s (set to %d) can be max %d", CfgMigratorReceiptMaxEntries, maxReceiptEntries, iotago.MaxMigratedFundsEntryCount))
+		case maxReceiptEntries == 0:
```
muXxer: why not `<= 0`

muXxer (approved):

plugins/migrator/plugin.go
Wollac: Can't we make the flag a `unit`?

Wollac (approved):
Alternatively the flag could be a `uint` but its essentially the same, so I am happy with it

PR #974, opened 2021-03-18
database/caching improvements

### LINKED ISSUE #914: High disc usage for no apparent reason
The node suddenly has a high disc usage. It is not clear to me when or why this happened.
The dashboard shows that the database size fluctuates between 15 and 25 GB within a few minutes.
Pruning and spamming are not active.

**To Reproduce**
No idea

**Environment information:**
 Ubuntu 20.04
 Hetzner CPX31 (Hornet only) 
"engine": "pebble",

Hornet Build 1d5576e

Stack-Dump: https://ybin.me/p/8175554e19c667e4#gTF5smDiVV8reekd7gtV1Ie0Jw9KE4E16uKw8FY2zuk=
[memory_profile.pdf](https://github.com/gohornet/hornet/files/6060026/memory_profile.pdf)
Bildschirmfoto 2021-03-01 um 08 22 20

muXxer:
Fixes #914


luca-moser (approved):
I'm for merging this even though I can't really say what effects the new changes to the Pebble settings will have.

muXxer:
Just for the reference here as a short explanation.

I tried to stick to the pebble defaults in most of the cases, and added all the comments from pebble so we can directly see what these parameters are for.

The moment we added the "compaction" stats to prometheus, we saw that after some time the node starts a compaction run, which somehow never stops. This leads to massive performance losses, since the disc IO and CPU usage is almost at maximum all the time. You can also see that described in the issue #914.

I was wondering what causes this, since we were using default settings. It seems that the read amplified triggered compactions cause this. Maybe a compaction also causes "read access" which then causes an compaction again, so we loop infinitely. After changing the values that cause "read amplified triggered" compactions to different values, the infinite compaction runs stopped. Now we see only small compaction runs, which seem to be triggered by "write amplification" only. I'm not a database expert, but I guess this is enough for our use case. We shouldn't encounter write stalls, because we still have write amplification triggered compactions.

PR #990, opened 2021-03-25
Use msg id for minted output msg metadata pointer

luca-moser:
We used the ms id as the msg id for how we store outputs. With this change, one can actually point back to the actual message which carried the milestone to get all the data back.


muXxer:
But this means that this info could be different in every node?
You did it on purpose on the milestoneID before, because the milestone payload can be reattached....

luca-moser:
Of course but I don't think it has any real implications in this context. Or do any spring to your mind?

muXxer:
If you query the same node for the message, which you queried for the treasury output, you will be fine.
Just wanted to mention :scam:

muXxer (approved):

luca-moser:
carlos

PR #994, opened 2021-03-28
Remove special case for broadcasting milestones

muXxer:
This PR removes the special handling for broadcasting milestones in "gossip-on-solidification".

It was requested by @luca-moser and @Wollac


luca-moser (approved):

GalRogozinski:
Why was it removed though?
I assume you can't know if a milestone is truly valid unless it is solid, but no info is given in the PR

PR #996, opened 2021-03-28
Fix/db revalidation

muXxer:
This PR fixes several problems with the database revalidation.

Database revalidation is not started automatically at node startup anymore.
The old behavior can still be achieved by enabling `db.autoRevalidation` in the config.
It also adds new flags to delete the database (`--deleteDatabase`) and snapshots (`--deleteAll`) at startup.


pkg/tangle/revalidation.go
```diff
@@ -125,6 +124,20 @@ func (t *Tangle) RevalidateDatabase() error {
 	return nil
 }
 
+// deletes all ledger entries (unspent, spent, diffs, balances, treasury, receipts).
+func (t *Tangle) cleanupLedger(info *storage.SnapshotInfo, pruneReceipts bool) error {
```
alexsporn: snapshotInfo is unused

alexsporn (approved):

PR #999, opened 2021-03-30
Add debug metrics to prometheus (snapshot/pruning/confirmation)

muXxer:


plugins/prometheus/debug.go
```diff
@@ -0,0 +1,156 @@
...
+	snapshotTotalDuration              prometheus.Histogram
+	snapshotDurations                  *prometheus.GaugeVec
+	databasePruningTotalDuration       prometheus.Histogram
+	databasePruningDurations           *prometheus.GaugeVec
+	milestoneConfirmationTotalDuration prometheus.Histogram
+	milestoneConfirmationDurations     *prometheus.GaugeVec
+
+	lastSnapshotMetricsPointer        unsafe.Pointer
+	lastDatabasePruningMetricsPointer unsafe.Pointer
+	lastConfirmationMetricsPointer    unsafe.Pointer
```
luca-moser: Remove the `unsafe.Pointer`s and use locks or anything else to deal with the data race.

luca-moser (changes requested):

plugins/prometheus/debug.go
muXxer: Fixed.

luca-moser (approved):

PR #1014, opened 2021-04-08
Choose snapshot type based on file size of last snapshots

### LINKED ISSUE #987: Decide whether to create a full or delta snapshot based on size
We implemented the possibility to create delta snapshots in chrysalis pt2 to reduce the needed download size for new snapshots.

Since delta snapshots always contain all milestones since the last full snapshot, the file size keeps growing.
At some point it could be larger than the full snapshot, which makes no sense at all.

We should add a config option to decide whether a delta snapshot should be taken or a full snapshot.
For example if the last delta snapshot is bigger than X% of the last full snapshot, we could trigger the full snapshot.

muXxer:
Fixes #987 

The new concept of full and delta snapshots, introduced in chrysalis, is used to reduce the download size of snapshot files.
A full snapshot file always contains the complete ledger state, while a delta snapshot file only contains the milestones including ledger changes since the last full snapshot file.

Creating and applying a delta snapshot file creates more IO pressure on the node, and the file size of a delta snapshot file can grow bigger than the last full snapshot file. Because of these reasons, it makes sense to create a full snapshot file instead of a delta snapshot file, if a certain ratio in the file size of the last snapshots is reached.

### DIFF
--- a/alphanet/config_alphanet.json
+++ b/alphanet/config_alphanet.json
@@ -69,6 +69,7 @@
     "interval": 200,
     "fullPath": "snapshots/alphanet/full_export.bin",
     "deltaPath": "snapshots/alphanet/delta_export.bin",
+    "deltaSizeThresholdPercentage": 50.0,
     "downloadURLs": []
   },
   "pruning": {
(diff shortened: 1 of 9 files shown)


luca-moser (changes requested):
Looks good. Add an exception where when `deltaSizeThresholdPercentage` is 0, it will always creates a delta snapshot. This is useful in case someone wants to ensure that the range between the full and delta snapshot always contains all the milestone diff data instead of it getting compressed.

luca-moser (approved):

PR #1023, opened 2021-04-11
Feat/future cone solidifier

muXxer:
This PR adds a future cone solidifier struct to avoid solidifications running in parallel.
It also improves the reuse of traversers and memcaches.

### DIFF
--- a/pkg/tangle/solidifier_future_cone.go
+++ b/pkg/tangle/solidifier_future_cone.go
@@ -1,0 +1,124 @@
+package tangle
+
+import (
+	"bytes"
+
+	"github.com/gohornet/hornet/pkg/dag"
+	"github.com/gohornet/hornet/pkg/model/hornet"
+	"github.com/gohornet/hornet/pkg/model/storage"
+	"github.com/iotaledger/hive.go/syncutils"
+)
+
+// FutureConeSolidifier traverses the future cone of messages and updates their solidify.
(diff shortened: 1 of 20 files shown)


pkg/tangle/solidifier_future_cone.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"bytes"
+
+	"github.com/gohornet/hornet/pkg/dag"
+	"github.com/gohornet/hornet/pkg/model/hornet"
+	"github.com/gohornet/hornet/pkg/model/storage"
+	"github.com/iotaledger/hive.go/syncutils"
+)
+
+// FutureConeSolidifier traverses the future cone of messages and updates their solidify.
```
luca-moser:
```suggestion
// FutureConeSolidifier traverses the future cone of messages and updates their solidity.
```

pkg/tangle/solidifier_future_cone.go
```diff
@@ -0,0 +1,124 @@
...
+// It holds a reference to a traverser and a memcache, so these can be reused for "gossip solidifcation".
```
luca-moser:
```suggestion
// It holds a reference to a traverser and a memcache, so that these can be reused for "gossip solidifcation".
```

luca-moser (approved):

PR #1044, opened 2021-04-25
Reset the HeaviestSelector if new milestone is confirmed.

muXxer:
This PR fixes a possible race condition in the COO, which could lead to an automatic shutdown of the COO node.

If a message gets solid and doesn't reference a cone that is below max depth, it is added to the heaviest selector.
Everytime a tipselection happens (checkpoints, milestone), this heaviest selector is resetted and all tracked messages are cleared.

There was a race condition, because messages are added to the heaviest selector between the tipselection for the milestone and the event that the milestone got confirmed in the COO node itself. This way messages could have been added to the tip pool that are already "below max depth". The selector is now resetted with the "onConfirmedMilestoneChanged" event and possible checkpoints are resetted as well.

### DIFF
--- a/plugins/coordinator/plugin.go
+++ b/plugins/coordinator/plugin.go
@@ -29,6 +29,7 @@ import (
 	"github.com/iotaledger/hive.go/configuration"
 	"github.com/iotaledger/hive.go/events"
 	"github.com/iotaledger/hive.go/logger"
+	"github.com/iotaledger/hive.go/syncutils"
 	"github.com/iotaledger/hive.go/timeutil"
 	"github.com/iotaledger/iota.go/v2/ed25519"
 )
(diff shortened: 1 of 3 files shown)


plugins/coordinator/plugin.go
```diff
@@ -498,6 +515,21 @@ func configureEvents() {
...
 
+	onConfirmedMilestoneIndexChanged = events.NewClosure(func(msIndex milestone.Index) {
+		heaviestSelectorLock.Lock()
+		defer heaviestSelectorLock.Unlock()
+
+		// the selector needs to be resetted after the milestone was confirmed, otherwise
+		// it could contain tips that are already below max depth.
+		deps.Selector.Reset()
+
+		// the checkpoint also needs to be resetted, otherwise
```
luca-moser:
```suggestion
		// the checkpoint also needs to be reset, otherwise
```

plugins/coordinator/plugin.go
```diff
@@ -498,6 +515,21 @@ func configureEvents() {
 		}
 	})
 
+	onConfirmedMilestoneIndexChanged = events.NewClosure(func(msIndex milestone.Index) {
+		heaviestSelectorLock.Lock()
+		defer heaviestSelectorLock.Unlock()
+
+		// the selector needs to be resetted after the milestone was confirmed, otherwise
```
luca-moser:
```suggestion
		// the selector needs to be reset after the milestone was confirmed, otherwise
```

luca-moser (approved):

PR #1085, opened 2021-05-05
Improve tools

### LINKED ISSUE #1073: Allow tools pwdhash to accept either stdin or ENV variable
Consider tooling like hornet-playbook that help to automate various node management processes, the ability to generate the salt+password hash automatically would be very handy.

Consider the following example to best illustrate the idea:
```sh
$ read -s MYPASS
<user inputs the password>
$ echo $MYPASS
somepassword
$ docker run -it --rm -e PWDHASH="$MYPASS" --name test gohornet/hornet:1.0.1  tools pwdhash --from-env
Success!
Your hash: 1b855eec4980d645fcdbedfd4ad06cccd51a49ddb5426b96e524ae9344be9051
Your salt: 39a2bc364bf5c08d8f4a2fd2a60c4327b581830a02c665d270611e9e4de12bcb
```
Maybe the `--from-env` or any other arg is not really necessary if the PWDHASH environment variable is detected as defined.

This way, it is possible to wrap around this functionality and automate the user's salt+password hash with e.g. scripts etc.

muXxer:
This PR adds the possibility to pass a password via environment variable, so the pwdhash tool can also be used by other applications directly.

EnvironmentVar `HORNET_TOOL_PASSWORD`

Fixes #1073

### DIFF
--- a/core/p2p/core.go
+++ b/core/p2p/core.go
@@ -3,32 +3,23 @@ package p2p
 
 import (
 	"context"
-	stded25519 "crypto/ed25519"
-	"encoding/hex"
+	"errors"
 	"fmt"
-	"io"
-	"io/ioutil"
-	"os"
 	"path"
-	"runtime"
 	"time"
 
-	badger "github.com/ipfs/go-ds-badger"
 	"github.com/libp2p/go-libp2p"
 	connmgr "github.com/libp2p/go-libp2p-connmgr"
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/host"
 	"github.com/libp2p/go-libp2p-core/peer"
-	"github.com/libp2p/go-libp2p-core/peerstore"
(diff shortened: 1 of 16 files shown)


core/p2p/core.go
```diff
@@ -224,6 +221,15 @@ func provide(c *dig.Container) {
 }
 
 func configure() {
+
+	// make sure nobody copies around the peer store since it contains the
+	// private key of the node
+	log.Infof("never share your %s folder as it contains your node's private key!", deps.NodeConfig.String(CfgP2PPeerStorePath))
+
+	if !peerStoreIsNew {
+		log.Infof("retrieving existing peer identity from %s", path.Join(deps.NodeConfig.String(CfgP2PPeerStorePath), pubKeyFileName))
```
luca-moser: Since this is now printed out on configure(), you can change it to be "retrieved existing peer identity..", since the initialization of the libp2p host already happened.

luca-moser (changes requested):

PR #1136, opened 2021-07-03
fix base58 pubkey size

no8ody:


Dr-Electron:
We could also remove the base58 length validation, as the byte validation (protoValidBytes) would fail. But in that case an entry node with an invalid base58 string length would lead to a panic instead of just a warning as it seems.
I will leave the decision to the hornet team ;)

luca-moser (approved):

luca-moser:
Thanks

PR #1157, opened 2021-08-13
Add faucet plugin

muXxer:
This PR adds the new faucet plugin.

Free coins for everyone :trollface:

### DIFF
--- a/config.json
+++ b/config.json
@@ -210,6 +210,16 @@
     "workers": 0,
     "autostart": false
   },
+  "faucet": {
+    "hrpNetworkPrefix": "atoi",
+    "amount": 10000000,
+    "smallAmount": 1000000,
+    "maxAddressBalance": 20000000,
+    "maxOutputCount": 127,
+    "indexationMessage": "HORNET FAUCET",
+    "batchTimeout": "2s",
+    "powWorkerCount": 0
+  },
   "mqtt": {
     "bindAddress": "localhost:1883",
     "wsPort": 1888,
--- a/config.json
+++ b/config.json
@@ -211,7 +211,6 @@
     "autostart": false
   },
   "faucet": {
-    "hrpNetworkPrefix": "atoi",
     "amount": 10000000,
     "smallAmount": 1000000,
     "maxAddressBalance": 20000000,
--- a/plugins/faucet/params.go
+++ b/plugins/faucet/params.go
@@ -10,9 +10,9 @@ import (
 )
 
 const (
-	// the amount of funds the requester receives if the target address has less funds than the given amount.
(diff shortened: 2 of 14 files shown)


config.json
```diff
@@ -210,6 +210,16 @@
     "workers": 0,
     "autostart": false
   },
+  "faucet": {
+    "hrpNetworkPrefix": "atoi",
```
alexsporn: this is wrong for config.json

plugins/faucet/params.go
```diff
@@ -0,0 +1,44 @@
...
+	CfgFaucetBatchTimeout = "faucet.batchTimeout"
+	// the amount of workers used for calculating PoW when issuing faucet messages.
+	CfgFaucetPoWWorkerCount = "faucet.powWorkerCount"
+)
+
+var params = &node.PluginParams{
+	Params: map[string]*flag.FlagSet{
+		"nodeConfig": func() *flag.FlagSet {
+			fs := flag.NewFlagSet("", flag.ContinueOnError)
+			fs.Int64(CfgFaucetAmount, 10000000, "the amount of funds the requester receives if the target address has less funds than the given amount")
```
alexsporn: `has less funds than the given amount` -> -`has less funds than the maximum allowed`

plugins/faucet/params.go
```diff
@@ -0,0 +1,44 @@
...
+
+const (
+	// the amount of funds the requester receives if the target address has less funds than the given amount.
+	CfgFaucetAmount = "faucet.amount"
+	// the amount of funds the requester receives if the target address has more funds than the given amount.
+	CfgFaucetSmallAmount = "faucet.smallAmount"
+	// the maximum allowed amount of funds on the target address.
+	CfgFaucetMaxAddressBalance = "faucet.maxAddressBalance"
+	// the maximum output count per faucet message.
+	CfgFaucetMaxOutputCount = "faucet.maxOutputCount"
```
alexsporn, line 20: why is this configurable? to avoid big pow?

muXxer: I can't remember but I just say yes :trollface:

plugins/faucet/params.go
muXxer: We have two thresholds, for example the "amount" is 10 Mi.

If someone has less than 10 Mi, he gets 10 Mi on his address.
If he has more than 10 Mi already, he only gets the small amount (1 Mi) until the max is reached (20 Mi).

alexsporn: I know, but your sentence in the help does not make sense. 
what is "the given amount" in that sentence?

alexsporn (approved):

PR #1186, opened 2021-10-01
Fix/warpsync milestone deadlock

muXxer:
This PR fixes an edge case, where the warpsync plugin could lead to a deadlock in the syncing process.

If for example the public keys of a milestone are not known to the node in the moment the milestone was received, the milestone message is stored to the database, but the payload is considered as invalid.

If the node operator then adds the correct public keys and the milestones are received again, the payload was not verified again because the message was already known to the node. This resulted in a deadlock in combination with the warpsync plugin, because warpsync will block the syncing process of newer milestones until the defined range of milestones are received and solidified.

The fix is to re-verify milestone payloads of unknown milestones, even if the message is already known.

### DIFF
--- a/pkg/protocol/gossip/rqueue.go
+++ b/pkg/protocol/gossip/rqueue.go
@@ -141,7 +141,8 @@ func (r *Request) MapKey() string {
 
 type Requests []*Request
 
-func (r Requests) Requested() bool {
+// HasRequest returns true if Requests contains a Request.
+func (r Requests) HasRequest() bool {
 	return len(r) > 0
 }
 
--- a/pkg/protocol/gossip/rqueue.go
+++ b/pkg/protocol/gossip/rqueue.go
@@ -2,14 +2,45 @@ package gossip
 
 import (
 	"container/heap"
+	"strconv"
 	"sync"
 	"time"
 
 	"go.uber.org/atomic"
 
+	"github.com/pkg/errors"
+
 	"github.com/gohornet/hornet/pkg/model/hornet"
 	"github.com/gohornet/hornet/pkg/model/milestone"
 )
+
+var (
+	ErrUnknownRequestType = errors.New("unknown request type")
+)
+
+// RequestType is the type of request.
+type RequestType int
+
+const (
+	RequestTypeMessageID RequestType = iota
+	RequestTypeMilestoneIndex
+)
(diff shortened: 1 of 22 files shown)


pkg/protocol/gossip/rqueue.go
```diff
@@ -86,6 +118,33 @@ type Request struct {
...
+	case RequestTypeMilestoneIndex:
+		return r.MilestoneIndex.String()
+	default:
+		panic(ErrUnknownRequestType)
+	}
+}
+
+type Requests []*Request
+
+func (r Requests) Requested() bool {
```
alexsporn: This logic is kind of weird. Requested is true if there are requests

pkg/protocol/gossip/warpsync.go
```diff
@@ -324,3 +324,46 @@ func (w *WarpSyncMilestoneRequester) Cleanup() {
...
 }
+
+// RequestMilestoneRange requests up to N milestones nearest to the current confirmed milestone index.
+// Returns the number of milestones requested.
+func (w *WarpSyncMilestoneRequester) RequestMilestoneRange(rangeToRequest int, onExistingMilestoneInRange func(milestone *storage.CachedMilestone), from ...milestone.Index) int {
+	var requested int
+
+	// make sure we only request what we don't have
+	startingPoint := w.syncManager.ConfirmedMilestoneIndex()
+	if len(from) > 0 {
```
alexsporn, line 334: compare from[0] to startingPoint?

pkg/protocol/gossip/warpsync.go
```diff
@@ -324,3 +324,46 @@ func (w *WarpSyncMilestoneRequester) Cleanup() {
...
+		}
+
+		cachedMs.Release(true) // milestone -1
+	}
+
+	if len(msIndexes) == 0 {
+		return requested
+	}
+
+	// enque every milestone request to the request queue
```
alexsporn: enqueue

pkg/protocol/gossip/warpsync.go
muXxer: I removed the confusing comment, but the logic will stay as it is for now.

alexsporn (approved):

PR #1191, opened 2021-10-11
Refactor the JWT auth for the API

### LINKED ISSUE #1072: JWT, permitted routes and making the node public.
**Is your feature request related to a problem? Please describe.**
Not sure whether this is a feature at this point, expected behavior or a bug.

**Context**
There are few issues here.
I think it is important to share what is being attempted first:

Many users want to have their nodes publicly available for Firefly, this requires TLS/HTTPS (done via a proxy e.g. nginx).
In addition, the users want to be able to manage things like peers via the dashboard.

I cannot configure nginx to restrict access to a route like `/api/v1/peers` (via e.g. auth basic) because that would prevent the dashboard's access to it.

One proposed idea was to enable JWT. However, I ran into some issues with it:

Dashboard cannot access API with JWT enabled
-----------------------------------------------------
Please note: this behavior was tested both with and without a proxy/nginx.
If JWT is enabled, the dashboard cannot work with the API any longer, I get the error:
`Failed to add peer: invalid jwt claims, error: code=401, message=invalid jwt claims`
I can access using e.g. curl only when setting to the JWT to the one from the Hornet log.

Expected behavior: when JWT is enabled, the dashboard should still have its full access to the API.

(By the way, I have 127.0.0.1 and ::1 whitelisted as in the original config.json)

Permitted Routes
-------------------
When enabling JWT, all the API's routes, even those in the `permittedRoutes` require the token.
But suppose I only want to require JWT for the non-permitted routes, I still want to let Firefly connect.
It seems to me that `permittedRoutes` should always be permitted, regardless of whether JWT is enabled or not.

**Extra info**
Using the docker hornet image version 1.0.1 on Ubuntu 20.04

Thank you.

alexsporn:
This change removes the `whitelistedAddresses` and `permittedRoutes` configuration option and separates the endpoints into 2 groups:
- `publicRoutes`: the routes exposed without any need of authorization
- `protectedRoutes`: the routes exposed with JWT authorization

This means that JWT authorization is now enabled by default (configuration option to enable was removed).
The `publicRoutes` and `protectedRoutes` options now support wildcards (`*`) to simplify the lists.

All routes matching the `publicRoutes` list will be exposed publicly, even if they are also matched by the `protectedRoutes` list.

_**Note:** The dashboard accesses the API to provide its functionality, so the required routes will be exposed by the dashboard. The exposed routes by the dashboard will only accept JWT tokens issued for the dashboard, so do not try to use the API tokens on the dashboard routes._

Fixes #1072


private_tangle/config_private_tangle.json
```diff
@@ -2,47 +2,13 @@
...
-      "/api/plugins/debug/requests",
-      "/api/plugins/debug/message-cones/*",
-      "/api/plugins/faucet/info",
-      "/api/plugins/faucet/enqueue"
-    ],
-    "whitelistedAddresses": [
-      "127.0.0.1",
-      "::1"
+      "/api/v1/*",
+      "/api/plugins/*"
```
muXxer, line 11: maybe add an empty `protectedRoutes` list here as well, because the private tangle config contains all possible config keys.

plugins/restapi/api.go
```diff
@@ -1,37 +1,171 @@
...
+		regexes = append(regexes, reg)
 	}
-	return false
+	return regexes
 }
 
-func middlewareFilterRoutes(whitelistedNetworks []*net.IPNet, permittedRoutes map[string]struct{}) echo.MiddlewareFunc {
+func apiMiddleware() echo.MiddlewareFunc {
+
+	publicRoutes := compileRoutesAsRegexes(deps.NodeConfig.Strings(CfgRestAPIPublicRoutes))
```
muXxer, line 41: Should we maybe pass all used NodeConfig parameters to the `apiMiddleware` function, so we see which ones are used by just looking at `plugin.go`?

documentation/docs/post_installation/post_installation.md
```diff
@@ -117,11 +117,9 @@ You can find the HTTP REST API related options in the  `restAPI` section within
 ```json
   "restAPI": {
     "jwtAuth": {
-      "enabled": false,
       "salt": "HORNET"
     },
-    "excludeHealthCheckFromAuth": false,
-    "permittedRoutes": [
+    "publicRoutes": [
```
muXxer, line 122: Thats different to the default config :thinking:

muXxer (approved):

plugins/restapi/api.go
muXxer: Since it would be too much parameters, just ignore that comment.
```go
func apiMiddleware(publicRoutes []string, protectedRoutes []string, jwtAuthSalt string, nodeId string, nodePrivateKey crypto.PrivKey, dashboardUsername string) echo.MiddlewareFunc {
```

PR #1195, opened 2021-10-12
Use contexts to cancel instead of signal channels

muXxer:
This PR replaces most of the signal channels we used for cancellation or shutdown with contexts.

This is more go-idiomatic.

### DIFF
--- a/pkg/tangle/solidifier.go
+++ b/pkg/tangle/solidifier.go
@@ -1,6 +1,7 @@
 package tangle
 
 import (
+	"context"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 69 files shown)


pkg/tangle/solidifier.go
```diff
@@ -140,13 +142,25 @@ func (t *Tangle) SolidQueueCheck(
...
-	if t.signalChanMilestoneStopSolidification != nil {
-		close(t.signalChanMilestoneStopSolidification)
-		t.signalChanMilestoneStopSolidification = nil
+	t.milestoneSolidificationCtxLock.Lock()
+	defer t.milestoneSolidificationCtxLock.Unlock()
+
+	if t.milestoneSolidificationCancelFunc != nil {
+		// cancel ongoing solidifications
+		t.milestoneSolidificationCancelFunc()
+		t.milestoneSolidificationCancelFunc = nil
```
alexsporn, line 162: `A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing`
Not sure I like the context being a global state like this

muXxer: Yeah I also don't like that whole solidification trigger/cancelation part tbh.
But this here should be a 1 to 1 replacement for the current solution. 
The signal channel was also used in the same "global state" way.

Let's refactor this another time maybe?

alexsporn: Ok, fine by me

alexsporn (approved):

PR #1197, opened 2021-10-13
Only accept bech32 addresses with the correct prefix in the rest API

alexsporn:


plugins/restapi/v1/utxo.go
```diff
@@ -116,11 +116,15 @@ func balanceByBech32Address(c echo.Context) (*addressBalanceResponse, error) {
...
 	addressParam := strings.ToLower(c.Param(ParameterAddress))
 
-	_, bech32Address, err := iotago.ParseBech32(addressParam)
+	hrp, bech32Address, err := iotago.ParseBech32(addressParam)
 	if err != nil {
 		return nil, errors.WithMessagef(restapi.ErrInvalidParameter, "invalid address: %s, error: %s", addressParam, err)
 	}
 
+	if hrp != deps.Bech32HRP {
+		return nil, errors.WithMessagef(restapi.ErrInvalidParameter, "Invalid bech32 address provided! Address does not start with \"%s\".", deps.Bech32HRP)
```
muXxer: The other error messages in that plugin are differently cased.
We only changed the format in the faucet because it is directly shown on the faucet website. :)

plugins/restapi/v1/utxo.go
```diff
@@ -233,11 +237,15 @@ func outputsIDsByBech32Address(c echo.Context) (*addressOutputsResponse, error)
...
 
 	addressParam := strings.ToLower(c.Param(ParameterAddress))
-	_, bech32Address, err := iotago.ParseBech32(addressParam)
+	hrp, bech32Address, err := iotago.ParseBech32(addressParam)
 	if err != nil {
 		return nil, errors.WithMessagef(restapi.ErrInvalidParameter, "invalid address: %s, error: %s", addressParam, err)
 	}
 
+	if hrp != deps.Bech32HRP {
+		return nil, errors.WithMessagef(restapi.ErrInvalidParameter, "Invalid bech32 address provided! Address does not start with \"%s\".", deps.Bech32HRP)
```
muXxer: same here

muXxer (approved):

Issue #1095, opened 2021-05-12
JSON option for cli tools output


nuriel77:
**Is your feature request related to a problem? Please describe.**
Not related to any problem.

**Describe the solution you'd like**

Would be handy to be able to request JSON output for cli tools commands, e.g.:

```
hornet tools p2pidentity --output json
```
Or (once it accepts password through ENV variable):
```
hornet tools pwdhash --output json
```
(or argument could just be `--json` if there's no plan/need to support other types in the future)

**Describe alternatives you've considered**
Currently parsing output via bash / sed / awk etc, but that is inconsistent and there's a chance output may change in the future  breaking tools that rely on it.

muXxer:
https://media.giphy.com/media/eaECZB7V6GACc/giphy.gif

nuriel77:
> https://media.giphy.com/media/eaECZB7V6GACc/giphy.gif

hahaha, no rush about this one @muXxer its just a nice-to-have ;)

PR #1199, opened 2021-10-14
Add JSON output to some of the tools

alexsporn:
Implements #1095

### DIFF
--- a/core/app/app.go
+++ b/core/app/app.go
@@ -82,6 +82,9 @@ var (
 )
 
 func initialize(params map[string][]*flag.FlagSet, maskedKeys []string) (*node.InitConfig, error) {
+
+	toolsRequested := toolset.ShouldHandleTools()
+
 	flagSets, err := normalizeFlagSets(params)
 	if err != nil {
 		return nil, err
@@ -97,8 +100,11 @@ Command line flags:
 		flag.PrintDefaults()
 	}
 
-	parseFlags(flagSets)
-	printVersion(flagSets)
+	if !toolsRequested {
+		// No tools requested, so parse all flags as config flags
+		parseFlags(flagSets)
+		printVersion(flagSets)
+	}
 
 	if err = loadCfg(flagSets); err != nil {
 		return nil, err
@@ -112,7 +118,9 @@ Command line flags:
 		panic(err)
 	}
 
-	toolset.HandleTools(nodeConfig)
+	if toolsRequested {
+		toolset.HandleTools(nodeConfig)
+	}
 
 	fmt.Printf(`
               ██╗  ██╗ ██████╗ ██████╗ ███╗   ██╗███████╗████████╗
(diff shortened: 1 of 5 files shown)


pkg/toolset/toolset.go
```diff
@@ -26,24 +26,24 @@ const (
...
+func ShouldHandleTools() bool {
 	args := os.Args[1:]
 
-	toolFound := false
 	for i, arg := range args {
 		if strings.ToLower(arg) == "tool" || strings.ToLower(arg) == "tools" {
 			args = args[i:]
-			toolFound = true
+			return true
 			break
```
muXxer: no need for the break anymore

core/app/app.go
```diff
@@ -97,8 +100,11 @@ Command line flags:
 		flag.PrintDefaults()
 	}
 
-	parseFlags(flagSets)
-	printVersion(flagSets)
+	if !toolsRequested {
+		// No tools requested, so parse all flags as config flags
+		parseFlags(flagSets)
```
muXxer: Wouldn't that imply that we can't pass the correct config to the tools anymore?
For example I want to run the tools with a config_devnet.json, I can't call `-c config_devnet.json` anymore?

The config is for example needed for the tools ToolP2PIdentityGen, ToolP2PExtractIdentity, ToolJWTApi.

pkg/toolset/ed25519.go
```diff
@@ -2,68 +2,115 @@ package toolset
...
 )
 
+type keys struct {
+	PublicKey      string `json:"publicKey"`
+	PrivateKey     string `json:"privateKey,omitempty"`
+	Ed25519Address string `json:"ed25519"`
+	Bech32Address  string `json:"bech32"`
+}
+
+func print(pubKey ed25519.PublicKey, privKey ed25519.PrivateKey, hrp iotago.NetworkPrefix, outputJSON bool) {
```
muXxer: Maybe rename to `printEd25519Info` ? :)

pkg/toolset/jwt.go
```diff
@@ -17,18 +19,23 @@ import (
...
-	printUsage := func() {
-		println("Usage:")
-		println(fmt.Sprintf("	%s [P2P_DATABASE_PATH]", ToolJWTApi))
-		println()
-		println("	[P2P_DATABASE_PATH] - the path to the p2p database folder (optional)")
-		println()
-		println(fmt.Sprintf("example: %s %s", ToolJWTApi, "p2pstore"))
+	fs := flag.NewFlagSet("", flag.ExitOnError)
+	outputJSON := fs.Bool("json", false, "format output as JSON")
+	databasePath := fs.String("database", "", "the path to the p2p database folder (optional)")
```
muXxer, line 24: maybe `database.path` ?
Or is this conflicting to our default setting?

alexsporn: why use long flags for the tools? They are long for the config, because we have a complex JSON structure

muXxer: ok

core/app/app.go
alexsporn: Ok, check again

muXxer (approved):

PR #1201, opened 2021-10-20
Separate UTXO database

alexsporn:
This PR splits the database into two separate ones.
- Tangle database, containing everything related to messages
- UTXO database, containing just the ledger and the ledger diffs

This split moves each database into a separate subdirectory inside the database dir.

The database will be automatically split when starting the node if required.

A new tool `db-split` was added to manually migrate a legacy database to the split format.

**_Notes:_**
- Prometheus database metrics now have both `tangle` and `utxo` metrics
- Dashboard metrics now have access to `tangle`, `utxo` and `total` sizes. A dashboard change is needed if we want to show these values separately

### DIFF
--- a/core/database/core.go
+++ b/core/database/core.go
@@ -28,6 +28,9 @@ const (
 	CfgTangleDeleteDatabase = "deleteDatabase"
 	// whether to delete the database and snapshots at startup
 	CfgTangleDeleteAll = "deleteAll"
+
+	TangleDatabaseDirectoryName = "tangle"
+	UTXODatabaseDirectoryName   = "utxo"
 )
 
 func init() {
@@ -93,8 +96,8 @@ func initConfigPars(c *dig.Container) {
 		return cfgResult{
 			DatabaseEngine:           dbEngine,
 			DatabasePath:             databasePath,
-			TangleDatabasePath:       filepath.Join(databasePath, "tangle"),
-			UTXODatabasePath:         filepath.Join(databasePath, "utxo"),
+			TangleDatabasePath:       filepath.Join(databasePath, TangleDatabaseDirectoryName),
+			UTXODatabasePath:         filepath.Join(databasePath, UTXODatabaseDirectoryName),
 			DeleteDatabaseFlag:       *deleteDatabase,
 			DeleteAllFlag:            *deleteAll,
 			DatabaseDebug:            deps.NodeConfig.Bool(CfgDatabaseDebug),
--- a/core/database/core.go
+++ b/core/database/core.go
@@ -202,19 +202,22 @@ func provide(c *dig.Container) {
 		Profile        *profile.Profile
 	}
 
-	if err := c.Provide(func(deps storageDeps) *storage.Storage {
+	type storageOut struct {
+		dig.Out
+		Storage     *storage.Storage
+		UTXOManager *utxo.Manager
+	}
+
+	if err := c.Provide(func(deps storageDeps) storageOut {
 
 		store, err := storage.New(deps.TangleDatabase.KVStore(), deps.UTXODatabase.KVStore(), deps.Profile.Caches)
 		if err != nil {
 			CorePlugin.Panicf("can't initialize storage: %s", err)
 		}
-		return store
-	}); err != nil {
-		CorePlugin.Panic(err)
-	}
-
-	if err := c.Provide(func(storage *storage.Storage) *utxo.Manager {
-		return storage.UTXOManager()
+		return storageOut{
+			Storage:     store,
+			UTXOManager: store.UTXOManager(),
+		}
 	}); err != nil {
 		CorePlugin.Panic(err)
 	}
--- a/core/database/core.go
+++ b/core/database/core.go
@@ -3,6 +3,7 @@ package database
 import (
 	"context"
 	"os"
+	"path/filepath"
 
 	flag "github.com/spf13/pflag"
 	"go.uber.org/dig"
@@ -58,7 +59,8 @@ var (
 
 type dependencies struct {
 	dig.In
-	Database       *database.Database
+	TangleDatabase *database.Database `name:"tangleDatabase"`
+	UTXODatabase   *database.Database `name:"utxoDatabase"`
 	Storage        *storage.Storage
 	StorageMetrics *metrics.StorageMetrics
 }
@@ -74,6 +76,8 @@ func initConfigPars(c *dig.Container) {
 		dig.Out
 		DatabaseEngine           database.Engine `name:"databaseEngine"`
 		DatabasePath             string          `name:"databasePath"`
+		TangleDatabasePath       string          `name:"tangleDatabasePath"`
+		UTXODatabasePath         string          `name:"utxoDatabasePath"`
 		DeleteDatabaseFlag       bool            `name:"deleteDatabase"`
 		DeleteAllFlag            bool            `name:"deleteAll"`
 		DatabaseDebug            bool            `name:"databaseDebug"`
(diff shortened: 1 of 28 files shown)


pkg/toolset/snap_hash.go
```diff
@@ -58,14 +59,23 @@ func snapshotHash(_ *configuration.Configuration, args []string) error {
 		return fmt.Errorf("database initialization failed: %w", err)
 	}
 
+	utxoStore, err := database.StoreWithDefaultSettings(filepath.Join(tempDir, "utxo"), true, targetEngine)
```
muXxer: should the other `store` be renamed to `tangleStore` and the dir also point to `tangle` subdir?

core/database/core.go
```diff
@@ -309,11 +296,19 @@ func run() {
 
 func closeDatabases() error {
 
-	if err := deps.Database.KVStore().Flush(); err != nil {
+	if err := deps.TangleDatabase.KVStore().Flush(); err != nil {
```
muXxer: Should maybe the `deps.Storage` have this flush logic for all DBs?

pkg/common/database_prefixes.go
```diff
@@ -1,14 +1,13 @@
 package common
 
 const (
-	StorePrefixHealth               byte = 0
+	StorePrefixHealthDeprecated     byte = 0
```
muXxer: Can you add a TODO to the `DbVersion` that we need to remove the Deprecated in the next datbase version update?

pkg/snapshot/snapshot_write.go
```diff
@@ -673,14 +674,23 @@ func MergeSnapshotsFiles(fullPath string, deltaPath string, targetFileName strin
...
 	}
 
+	utxoStore, err := database.StoreWithDefaultSettings(filepath.Join(tempDir, "utxo"), true, targetEngine)
+	if err != nil {
+		return nil, fmt.Errorf("utxo database initialization failed: %w", err)
+	}
+
 	// clean up temp db
 	defer func() {
 		store.Shutdown()
```
muXxer: maybe also rename to tangleDB?

pkg/snapshot/snapshot_write.go
```diff
@@ -673,14 +674,23 @@ func MergeSnapshotsFiles(fullPath string, deltaPath string, targetFileName strin
 		return nil, fmt.Errorf("database initialization failed: %w", err)
```
muXxer: Isn't the subfolder for tangle missing here?

pkg/testsuite/test_environment.go
```diff
@@ -59,16 +59,19 @@ type TestEnvironment struct {
...
+	utxoStore kvstore.KVStore
 
 	// storage is the tangle storage for this test.
 	storage *storage.Storage
 
 	// syncManager is used to determine the sync status of the node in this test.
 	syncManager *syncmanager.SyncManager
 
-	// milestoneManager is used to retrieve, verify and store milestones.
+	// milestoneManager is used to retrieve, verify and tangleStore milestones.
```
muXxer: tangleStore? lel :D
it should remain "store" in this context :)

pkg/toolset/coordinator_fix_state.go
```diff
@@ -44,13 +45,21 @@ func coordinatorFixStateFile(_ *configuration.Configuration, args []string) erro
 		return fmt.Errorf("database initialization failed: %w", err)
```
muXxer: same here, tangleStore and subdir?

pkg/snapshot/pruning.go
```diff
@@ -30,19 +30,26 @@ func (s *SnapshotManager) calcTargetIndexBySize(targetSizeBytes ...int64) (miles
...
 	if err != nil {
 		return 0, err
 	}
 
+	currentUTXODatabaseSizeBytes, err := s.utxoDatabase.Size()
+	if err != nil {
+		return 0, err
+	}
+
+	currentDatabaseSizeBytes := currentTangleDatabaseSizeBytes + currentUTXODatabaseSizeBytes
```
muXxer, line 51: the question is... should we somehow explain to the user that only the tangleDatabase can be pruned?
Maybe also only prune based on that size?
Or does it make no sense from a node maintanance view? I guess not :shrug:

pkg/model/storage/health_db.go
```diff
@@ -1,101 +1,103 @@
...
-	return s.healthStore.Flush()
-}
-
-func (s *Storage) MarkDatabaseTainted() error {
+func (s *Storage) MarkDatabasesCorrupted() error {
 
-	if err := s.healthStore.Set([]byte("dbTainted"), []byte{}); err != nil {
-		return errors.Wrap(NewDatabaseError(err), "failed to set database health status")
+	for _, h := range s.healthTrackers {
+		err := h.markCorrupted()
```
muXxer: Shouldn't we try to mark both as corrupted, even if the first one fails, and then return the error?

pkg/model/storage/health_db.go
```diff
@@ -1,101 +1,103 @@
...
+	return nil
 }
 
-func (s *Storage) MarkDatabaseHealthy() error {
+func (s *Storage) MarkDatabasesTainted() error {
 
-	if err := s.healthStore.Delete([]byte("dbCorrupted")); err != nil {
-		return errors.Wrap(NewDatabaseError(err), "failed to set database health status")
+	for _, h := range s.healthTrackers {
+		err := h.markTainted()
```
muXxer: Same here.

pkg/model/storage/health_db.go
```diff
@@ -1,101 +1,103 @@
...
 }
 
-func (s *Storage) IsDatabaseCorrupted() (bool, error) {
+func (s *Storage) MarkDatabasesHealthy() error {
 
-	contains, err := s.healthStore.Has([]byte("dbCorrupted"))
-	if err != nil {
-		return true, errors.Wrap(NewDatabaseError(err), "failed to read database health status")
+	for _, h := range s.healthTrackers {
+		err := h.markHealthy()
```
muXxer: Why not in a single line?
`if err := h.markHealthy(); err != nil {`

pkg/model/storage/storage.go
```diff
@@ -49,19 +49,22 @@ type Storage struct {
...
-func New(store kvstore.KVStore, cachesProfile ...*profile.Caches) (*Storage, error) {
+func New(store kvstore.KVStore, utxoStore kvstore.KVStore, cachesProfile ...*profile.Caches) (*Storage, error) {
 
-	utxoManager := utxo.New(store)
+	utxoManager := utxo.New(utxoStore)
 
 	s := &Storage{
-		store:       store,
+		healthTrackers: []*storeHealthTracker{
+			newStoreHealthTracker(store),
```
muXxer: maybe rename store to tangleStore?

pkg/toolset/database_split.go
```diff
@@ -0,0 +1,46 @@
...
+	flag "github.com/spf13/pflag"
+
+	"github.com/gohornet/hornet/core/database"
+	"github.com/iotaledger/hive.go/configuration"
+)
+
+func databaseSplit(_ *configuration.Configuration, args []string) error {
+
+	fs := flag.NewFlagSet("", flag.ExitOnError)
+	databasePath := fs.String("database", "", "the path to the p2p database folder")
```
muXxer: p2p database? no...

plugins/prometheus/database.go
```diff
@@ -1,100 +1,132 @@
...
+		})
+
+	m.compactionCount = prometheus.NewCounter(
 		prometheus.CounterOpts{
 			Namespace: "iota",
 			Subsystem: "database",
-			Name:      "pruning_count",
-			Help:      "The total amount of database prunings.",
+			Name:      fmt.Sprintf("%s_compaction_count", name),
+			Help:      "The total amount of tangle database compactions.",
```
muXxer: `tangle` should be replaced by `name`

pkg/common/database_prefixes.go
alexsporn: I'll move it to the file holding the splitting code, so we can remove it there as soon as we remove any splitting code

pkg/snapshot/pruning.go
alexsporn: we do prune the UTXO one, we remove the "spent" UTXOs from it

muXxer (approved):

PR #1222, opened 2021-11-30
Fix/faucet conflicting tx

muXxer:
This PR fixes the broken faucet logic.

The transaction chaining logic, which is used to guarantee high throughput of the faucet, deadlocked in some edge cases, resulting in permanent double spents.

The new logic is based on the milestone ApplyConfirmation event.

### DIFF
--- a/pkg/model/faucet/faucet.go
+++ b/pkg/model/faucet/faucet.go
@@ -19,6 +19,8 @@ import (
 	"github.com/gohornet/hornet/pkg/model/utxo"
 	"github.com/gohornet/hornet/pkg/pow"
 	"github.com/gohornet/hornet/pkg/restapi"
+	"github.com/gohornet/hornet/pkg/whiteflag"
+	"github.com/iotaledger/hive.go/daemon"
 	"github.com/iotaledger/hive.go/events"
 	"github.com/iotaledger/hive.go/logger"
 	"github.com/iotaledger/hive.go/serializer"
@@ -35,6 +37,8 @@ type TipselFunc = func() (tips hornet.MessageIDs, err error)
 var (
 	// ErrNoTipsGiven is returned when no tips were given to issue a message.
 	ErrNoTipsGiven = errors.New("no tips given")
+	// ErrNothingToProcess is returned when there is no need to sweep or send funds.
+	ErrNothingToProcess = errors.New("nothing to process")
 )
 
 // Events are the events issued by the faucet.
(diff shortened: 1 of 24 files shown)


pkg/model/faucet/faucet.go
```diff
@@ -562,10 +594,101 @@ func (f *Faucet) logSoftError(err error) {
...
+}
+
+// processRequestsWithoutLocking processes all possible requests considering the maximum transaction size and the remaining funds of the faucet.
+// write lock must be acquired outside.
+func (f *Faucet) processRequestsWithoutLocking(collectedRequestsCounter int, amount uint64, batchedRequests []*queueItem) []*queueItem {
+	processedBatchedRequests := []*queueItem{}
+	unprocessedBatchedRequests := []*queueItem{}
+	nodeAlmostSynced := f.syncManager.IsNodeAlmostSynced()
+
+	for i := 0; i < len(batchedRequests); i++ {
```
alexsporn: why not use range iteration?

pkg/model/faucet/test/testenv.go
```diff
@@ -0,0 +1,353 @@
...
+	case <-chanDone:
+	case <-time.After(1 * time.Second):
+		env.t.Error("attachment of faucet message took too long")
+	}
+
+	return tips, nil
+}
+
+// RequestFunds sends requests to the faucet and waits until the next faucet message is issued.
+func (env *FaucetTestEnv) RequestFunds(wallets []*utils.HDWallet) (hornet.MessageIDs, error) {
```
alexsporn: if you make wallets variadic, then you don't need to always wrap into an array when you want to call it with just 1

pkg/model/faucet/faucet_test.go
```diff
@@ -0,0 +1,425 @@
...
+
+	// max reached
+	err = env.RequestFundsAndIssueMilestone([]*utils.HDWallet{env.Wallet1})
+	require.Error(t, err)
+}
+
+func TestMultipleRequests(t *testing.T) {
+	// requests to multiple addresses
+
+	var faucetBalance uint64 = 1000000000         //  1 Gi
```
alexsporn: It is easier to read if you group the digits with `_`, e.g. `1_000_000_000`

pkg/model/faucet/faucet_test.go
```diff
@@ -0,0 +1,425 @@
...
+		LatestMilestonesAsParents().
+		FromWallet(env.GenesisWallet).
+		ToWallet(env.FaucetWallet).
+		Amount(faucetAmount).
+		Build().
+		Store().
+		BookOnWallets()
+
+	// Confirming milestone at message
+	_, _ = env.IssueMilestone(hornet.MessageIDs{message.StoredMessageID()}...)
```
alexsporn: why are you wrapping a messageId in an array and then passing it as variadic? just pass `message.StoredMessageID()`

alexsporn (approved):

PR #1230, opened 2021-12-08
Stardust ✨

alexsporn:
✨✨✨✨


private_tangle/config_private_tangle.json
```diff
@@ -162,6 +162,7 @@
     "profile": "auto",
     "disablePlugins": [],
     "enablePlugins": [
+      "Indexer",
```
muXxer, line 168: vByteCosts are missing!

pkg/model/participation/event.go
```diff
@@ -89,15 +104,7 @@ func (e *Event) Deserialize(data []byte, deSeriMode serializer.DeSerializationMo
 		ReadNum(&e.MilestoneIndexEnd, func(err error) error {
 			return fmt.Errorf("unable to deserialize event end milestone: %w", err)
 		}).
-		ReadPayload(func(seri serializer.Serializable) { e.Payload = seri }, deSeriMode, func(ty uint32) (serializer.Serializable, error) {
-			switch ty {
-			case BallotPayloadTypeID:
```
muXxer, line 94: I can't see how this check is done now....
We only pass the PayloadSelector by passing the ReadGuard as far as I can see.

pkg/model/participation/participation_manager.go
```diff
@@ -752,15 +770,16 @@ func (pm *ParticipationManager) ParticipationsFromMessage(msg *storage.Message)
 		return nil, nil, nil
 	}
 
-	// only OutputSigLockedSingleOutput and OutputSigLockedDustAllowanceOutput are allowed as output type
-	switch depositOutputs[0].OutputType() {
-	case iotago.OutputSigLockedDustAllowanceOutput:
-	case iotago.OutputSigLockedSingleOutput:
+	// only OutputExtended are allowed as output type
```
muXxer:
```suggestion
	// only ExtendedOutput are allowed as output type
```

pkg/model/participation/participations.go
```diff
@@ -18,76 +16,67 @@ var (
 	participationsArrayRules = &serializer.ArrayRules{
 		Min:            ParticipationsMinParticipationCount,
 		Max:            ParticipationsMaxParticipationCount,
-		ValidationMode: serializer.ArrayValidationModeNone,
+		ValidationMode: serializer.ArrayValidationModeNoDuplicates,
+		UniquenessSliceFunc: func(next []byte) []byte {
+			return next[:EventIDLength] // Verify that the EventIDs are unique
```
muXxer, line 21: Is it guaranteed that `next` is at least `EventIDLength` in size?

pkg/model/participation/question.go
```diff
@@ -21,93 +19,78 @@ var (
 	answersArrayRules = &serializer.ArrayRules{
 		Min:            QuestionMinAnswersCount,
 		Max:            QuestionMaxAnswersCount,
-		ValidationMode: serializer.ArrayValidationModeNone,
+		ValidationMode: serializer.ArrayValidationModeNoDuplicates,
+		UniquenessSliceFunc: func(next []byte) []byte {
+			return next[:1] // Check answer value for uniqueness
```
muXxer, line 24: Is it guaranteed that `next` is at least `1` in size?

pkg/model/utxo/receipt.go
```diff
@@ -163,21 +163,27 @@ func (u *Manager) ForEachReceiptTupleMigratedAt(migratedAtIndex milestone.Index,
...
 	}
 	return outputs, nil
 }
 
 // OutputIDForMigratedFunds returns the UTXO ID for a migrated funds entry given the milestone containing the receipt
 // and the index of the entry.
-func OutputIDForMigratedFunds(milestoneHash iotago.MilestoneID, outputIndex uint16) iotago.UTXOInputID {
-	var utxoID iotago.UTXOInputID
+func OutputIDForMigratedFunds(milestoneHash iotago.MilestoneID, outputIndex uint16) iotago.OutputID {
+	var utxoID iotago.OutputID
```
muXxer: Maybe also rename `utxoID` here to `outputID` like above.

tools/participation-cli/main.go
```diff
@@ -67,52 +73,38 @@ func buildTransactionPayload(ctx context.Context, client *iotago.NodeHTTPAPIClie
...
-		balance, err := unspentOutput.Deposit()
-		if err != nil {
-			return nil, err
-		}
-
-		inputsBalance += int(balance)
+		inputsBalance += unspentOutput.Deposit()
 		txBuilder.AddInput(&iotago.ToBeSignedUTXOInput{Address: inputAddress, Input: input})
-
-		if inputsBalance >= (int(outputAmount) + MINIMUM_DUST_BALANCE) {
```
muXxer, line 82: It's wrong to remove this check, since you will always collect all UTXOs.
It should be this instead:
```diff
-		if inputsBalance >= (int(outputAmount) + MINIMUM_DUST_BALANCE) {
+		if inputsBalance >= int(outputAmount) {
```

tools/participation-cli/main.go
```diff
@@ -67,52 +73,38 @@ func buildTransactionPayload(ctx context.Context, client *iotago.NodeHTTPAPIClie
...
-	txBuilder.AddOutput(&iotago.SigLockedSingleOutput{Address: outputAddress, Amount: outputAmount})
-	inputsBalance -= int(outputAmount)
-
-	if inputsBalance != 0 && inputsBalance < MINIMUM_DUST_BALANCE {
-		return nil, fmt.Errorf("remainder does not fulfill the minimum balance requirement: %d, needed: %d", inputsBalance, MINIMUM_DUST_BALANCE)
-	}
+	txBuilder.AddOutput(&iotago.ExtendedOutput{Address: outputAddress, Amount: outputAmount})
+	inputsBalance -= outputAmount
 
 	if inputsBalance > 0 {
```
muXxer, line 111: Is there a way to determine the byte costs for an Extended output?
If yes, the remainder should be added to the output containing the participation if the remainder is less than the byte costs for the Extended output we create for the remainder.

tools/intsnap/main.go
```diff
@@ -52,106 +61,34 @@ var fullSnapshotHeader = &snapshot.FileHeader{
...
-			},
-			{
-				MessageID:  static32ByteID(3),
-				OutputID:   static34ByteID(3),
-				OutputType: iotago.OutputSigLockedSingleOutput,
-				Address:    staticEd25519Address(3),
-				Amount:     fullSnapshotOutputs[1].Amount,
-			},
+		Created: utxo.Outputs{
+			utxoOutput(3, fullSnapshotOutputs[0].Deposit()),
```
muXxer: Isn't the `3` and the `4` mixed here?

tools/intsnap/main.go
```diff
@@ -366,17 +249,39 @@ func static32ByteID(fill byte) [32]byte {
...
 	var addr iotago.Ed25519Address
 	copy(addr[:], b[:])
 	return &addr
 }
+
+func utxoOutput(fill byte, amount uint64) *utxo.Output {
+	return utxo.CreateOutput(
+		staticOutputID(fill),
+		staticMessageID(fill),
+		1,
```
muXxer: We don't care about the correct milestone it was created here?

plugins/restapi/v1/types.go
```diff
@@ -134,6 +134,11 @@ type OutputResponse struct {
 	OutputIndex uint16 `json:"outputIndex"`
 	// Whether this output is spent.
 	Spent bool `json:"isSpent"`
+	//TODO: expose the spent milestoneIndex and targetTransactionID
+	// The milestone index at which this output was booked into the ledger.
+	MilestoneIndex milestone.Index `json:"milestoneIndex"`
```
muXxer: Should we make it more clear by naming it `bookedMilestoneIndex` or `milestoneIndexBooked` ?

plugins/restapi/v1/types.go
```diff
@@ -134,6 +134,11 @@ type OutputResponse struct {
...
+	// The milestone timestamp this output was booked in the ledger.
+	MilestoneTimestamp uint32 `json:"milestoneTimestamp"`
```
muXxer: Should we make it more clear by naming it `bookedMilestoneTimestamp` or `milestoneTimestampBooked` ?

plugins/spammer/params.go
```diff
@@ -9,10 +9,10 @@ import (
 const (
 	// the message to embed within the spam messages
 	CfgSpammerMessage = "spammer.message"
-	// the indexation of the message
-	CfgSpammerIndex = "spammer.index"
-	// the indexation of the message if the semi-lazy pool is used (uses "index" if empty)
-	CfgSpammerIndexSemiLazy = "spammer.indexSemiLazy"
+	// the tag of the message
+	CfgSpammerTag = "spammer.tag"
+	// the tag of the message if the semi-lazy pool is used (uses "index" if empty)
```
muXxer:
```suggestion
	// the tag of the message if the semi-lazy pool is used (uses "tag" if empty)
```

plugins/spammer/params.go
```diff
@@ -28,8 +28,8 @@ var params = &node.PluginParams{
 		"nodeConfig": func() *flag.FlagSet {
 			fs := flag.NewFlagSet("", flag.ContinueOnError)
 			fs.String(CfgSpammerMessage, "IOTA - A new dawn", "the message to embed within the spam messages")
-			fs.String(CfgSpammerIndex, "HORNET Spammer", "the indexation of the message")
-			fs.String(CfgSpammerIndexSemiLazy, "HORNET Spammer Semi-Lazy", "the indexation of the message if the semi-lazy pool is used (uses \"index\" if empty)")
+			fs.String(CfgSpammerTag, "HORNET Spammer", "the tag of the message")
+			fs.String(CfgSpammerTagSemiLazy, "HORNET Spammer Semi-Lazy", "the tag of the message if the semi-lazy pool is used (uses \"index\" if empty)")
```
muXxer:
```suggestion
			fs.String(CfgSpammerTagSemiLazy, "HORNET Spammer Semi-Lazy", "the tag of the message if the semi-lazy pool is used (uses \"tag\" if empty)")
```

documentation/docs/post_installation/configuration.md
```diff
@@ -688,23 +688,23 @@ Example:
...
-| indexSemiLazy | The indexation of the message if the semi-lazy pool is used (uses "index" if empty) | string  |
-| cpuMaxUsage   | Workers remains idle for a while when cpu usage gets over this limit (0 = disable)  | float   |
-| mpsRateLimit  | The rate limit for the spammer (0 = no limit)                                       | float   |
-| workers       | The amount of parallel running spammers                                             | integer |
-| autostart     | Automatically start the spammer on node startup                                     | bool    |
+| Name         | Description                                                                        | Type    |
+|:-------------|:-----------------------------------------------------------------------------------| :------ |
+| message      | The message to embed within the spam messages                                      | string  |
+| tag          | The tag of the message                                                             | string  |
+| tagSemiLazy  | The tag of the message if the semi-lazy pool is used (uses "index" if empty)       | string  |
```
muXxer:
```suggestion
| tagSemiLazy  | The tag of the message if the semi-lazy pool is used (uses "tag" if empty)         | string  |
```

tools/intsnap/main.go
alexsporn: No, it is just metadata and is not used for any logic

tools/intsnap/main.go
alexsporn: You are right!

pkg/model/participation/participations.go
alexsporn: From what I understand from the serializer code it is first checked if the bytes can be parsed into the object, before the same bytes are used to pass to the UniquenessSliceFunc.

pkg/model/participation/question.go
alexsporn: From what I understand from the serializer code it is first checked if the bytes can be parsed into the object, before the same bytes are used to pass to the UniquenessSliceFunc.

pkg/model/participation/event.go
alexsporn: There is some magic in the serializer in the `readSerializableIntoTarget` func

pkg/indexer/alias.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"github.com/gohornet/hornet/pkg/model/milestone"
+	iotago "github.com/iotaledger/iota.go/v3"
+)
+
+type alias struct {
+	AliasID         aliasIDBytes  `gorm:"primaryKey;notnull"`
+	OutputID        outputIDBytes `gorm:"unique;notnull"`
+	Amount          uint64        `gorm:"notnull"`
+	StateController addressBytes  `gorm:"not null;index:alias_state_controller"`
```
muXxer:
```suggestion
	StateController addressBytes  `gorm:"notnull;index:alias_state_controller"`
```

pkg/indexer/alias.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"github.com/gohornet/hornet/pkg/model/milestone"
+	iotago "github.com/iotaledger/iota.go/v3"
+)
+
+type alias struct {
+	AliasID         aliasIDBytes    `gorm:"primaryKey;notnull"`
+	OutputID        outputIDBytes   `gorm:"unique;notnull"`
+	Amount          uint64          `gorm:"notnull"`
+	StateController addressBytes    `gorm:"not null;index:alias_state_controller"`
```
muXxer:
```suggestion
	StateController addressBytes    `gorm:"notnull;index:alias_state_controller"`
```

pkg/indexer/extended.go
```diff
@@ -0,0 +1,121 @@
...
+	if opts.sender != nil {
+		addr, err := addressBytesForAddress(*opts.sender)
+		if err != nil {
+			return errorResult(err)
+		}
+		query = query.Where("sender = ?", addr[:])
+	}
+
+	if opts.tag != nil && len(opts.tag) > 0 {
+		query = query.Where("tag = ?", opts.tag)
```
muXxer, line 117: Should we support prefix filtering?

pkg/indexer/extended.go
```diff
@@ -0,0 +1,121 @@
...
+		addr, err := addressBytesForAddress(*opts.unlockableByAddress)
+		if err != nil {
+			return errorResult(err)
+		}
+		query = query.Where("address = ?", addr[:])
+	}
+
+	if opts.requiresDustReturn != nil {
+		if *opts.requiresDustReturn {
+			query = query.Where("dust_return IS NOT NULL")
```
muXxer, line 102: Is this condition correct? Couldn't it be set but FALSE?

pkg/indexer/nft.go
```diff
@@ -0,0 +1,147 @@
...
+	return func(args *NFTFilterOptions) {
+		args.offset = offset
+	}
+}
+
+func nftFilterOptions(optionalOptions []NFTFilterOption) *NFTFilterOptions {
+	result := &NFTFilterOptions{
+		unlockableByAddress: nil,
+		requiresDustReturn:  nil,
+		sender:              nil,
```
muXxer, line 87: `issuer` is missing here

pkg/indexer/nft.go
```diff
@@ -0,0 +1,147 @@
...
+		addr, err := addressBytesForAddress(*opts.unlockableByAddress)
+		if err != nil {
+			return errorResult(err)
+		}
+		query = query.Where("address = ?", addr[:])
+	}
+
+	if opts.requiresDustReturn != nil {
+		if *opts.requiresDustReturn {
+			query = query.Where("dust_return > 0")
```
muXxer: Thats different to the query in the `extendedOutput`...

pkg/indexer/nft.go
```diff
@@ -0,0 +1,147 @@
...
+	if opts.sender != nil {
+		addr, err := addressBytesForAddress(*opts.sender)
+		if err != nil {
+			return errorResult(err)
+		}
+		query = query.Where("sender = ?", addr[:])
+	}
+
+	if opts.tag != nil && len(opts.tag) > 0 {
+		query = query.Where("tag = ?", opts.tag)
```
muXxer, line 144: Should we support prefix filtering?

pkg/indexer/types.go
```diff
@@ -0,0 +1,138 @@
...
+func (q queryResults) IDs() iotago.OutputIDs {
+	outputIDs := iotago.OutputIDs{}
+	for _, r := range q {
+		outputIDs = append(outputIDs, r.OutputID.ID())
+	}
+	return outputIDs
+}
+
+func (a addressBytes) Address() (iotago.Address, error) {
+	if len(a) != iotago.NFTAddressBytesLength || len(a) != iotago.AliasAddressBytesLength || len(a) != iotago.Ed25519AddressBytesLength {
```
muXxer:
```suggestion
	if len(a) != iotago.NFTAddressBytesLength && len(a) != iotago.AliasAddressBytesLength && len(a) != iotago.Ed25519AddressBytesLength {
```

pkg/indexer/types.go
```diff
@@ -0,0 +1,138 @@
...
+	return &IndexerResult{
+		Error: err,
+	}
+}
+
+func (i *Indexer) combineOutputIDFilteredQuery(query *gorm.DB, pageSize int, offset []byte) *IndexerResult {
+
+	query = query.Select("output_id", "milestone_index").Order("milestone_index asc, output_id asc")
+	if pageSize > 0 {
+		if offset != nil {
```
muXxer, line 79: Maybe we could support the offset query even if the pageSize is 0.

pkg/indexer/types.go
```diff
@@ -0,0 +1,138 @@
...
+			if len(offset) != OffsetLength {
+				return errorResult(errors.Errorf("Invalid offset length: %d", len(offset)))
+			}
+			msIndex := binary.LittleEndian.Uint32(offset[:4])
+			query = query.Where("milestone_index >= ?", msIndex).Where("output_id >= hex(?)", hex.EncodeToString(offset[4:]))
```
muXxer, line 84: We discussed that after coordicide this will be based on transaction timestamps instead of milestones.
With stardust we trust that milestones issued by the coordinator have a valid timestamp.
Maybe we can use that milestone timestamp here/everywhere already instead of the index?
This way nothing changes with coordicide here.

pkg/indexer/indexer.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+)
+
+type Indexer struct {
+	db *gorm.DB
+}
+
+func NewIndexer(dbPath string) (*Indexer, error) {
+
+	if err := utils.CreateDirectory(dbPath, 0700); err != nil {
```
muXxer, line 36: Should we also use our `CheckDatabaseEngine` function here, so we put the `dbInfo` file inside the folder, in case we support other engines for the indexer in the future as well?

pkg/indexer/indexer.go
```diff
@@ -0,0 +1,418 @@
...
+
+		if senderBlock := features.SenderFeatureBlock(); senderBlock != nil {
+			nft.Sender, err = addressBytesForAddress(senderBlock.Address)
+			if err != nil {
+				return err
+			}
+		}
+
+		if tagBlock := features.TagFeatureBlock(); tagBlock != nil {
+			copy(nft.Tag, tagBlock.Tag)
```
muXxer, line 249: Does this copy work if the `nft.Tag` is not initialized with the correct size?

pkg/indexer/indexer.go
```diff
@@ -0,0 +1,418 @@
...
+
+		if senderBlock := features.SenderFeatureBlock(); senderBlock != nil {
+			extended.Sender, err = addressBytesForAddress(senderBlock.Address)
+			if err != nil {
+				return err
+			}
+		}
+
+		if tagBlock := features.TagFeatureBlock(); tagBlock != nil {
+			copy(extended.Tag, tagBlock.Tag)
```
muXxer, line 100: Does this copy work if the `extended.Tag` is not initialized with the correct size?

alexsporn: I'll add `nft.Tag = make([]byte, len(tagBlock.Tag))` before the copy

pkg/indexer/extended.go
alexsporn: Don't understand you. We read the value of the bool pointer and decide
if true -> check IS NOT NULL
if false -> check IS NULL

pkg/indexer/types.go
alexsporn: I actually went ahead and removed it altogether, since it is unused

pkg/indexer/nft.go
alexsporn: Better add a ticket and add that after this PR gets merged

pkg/indexer/types.go
alexsporn: I think what you want is a different filter. You want anything newer than a certain output.
Better add a ticket and add that after this PR gets merged

pkg/indexer/indexer.go
alexsporn: Not sure this makes sense in this case. We are constructing queries that are not guaranteed to work with any other database engine right now.

integration-tests/tester/framework/config.go
```diff
@@ -33,6 +33,7 @@ const (
 
 	GenesisAddressPublicKeyHex = "f7868ab6bb55800b77b8b74191ad8285a9bf428ace579d541fda47661803ff44"
 	GenesisAddressHex          = "6920b176f613ec7be59e68fc68f597eb3393af80f74c7c3db78198147d5f1f92"
+	GensisAddressBech32        = "atoi1qp5jpvtk7cf7c7l9ne50c684jl4n8ya0srm5clpak7qes9ratu0eysflmsz"
```
muXxer:
```suggestion
	GenesisAddressBech32       = "atoi1qp5jpvtk7cf7c7l9ne50c684jl4n8ya0srm5clpak7qes9ratu0eysflmsz"
```

pkg/indexer/indexer.go
muXxer: Yeah, but this way it's consistent.
The file may also be parsed by whatever external tool.

plugins/participation/plugin.go
```diff
@@ -141,7 +143,7 @@ func provide(c *dig.Container) {
 }
 
 func configure() {
-	restapiv1.AddFeature(Plugin.Name)
+	restapiv2.AddPlugin("participation")
```
muXxer: Beginning with stardust, should we just version all plugins?
```suggestion
	restapiv2.AddPlugin("participation/v1")
```

plugins/mqtt/utils.go
```diff
@@ -219,13 +202,13 @@ func publishOutput(ledgerIndex milestone.Index, output *utxo.Output, spent bool)
 	outputsTopic := strings.ReplaceAll(topicOutputs, "{outputId}", output.OutputID().ToHex())
 	outputsTopicHasSubscribers := deps.MQTTBroker.HasSubscribers(outputsTopic)
 
-	addressBech32Topic := strings.ReplaceAll(topicAddressesOutput, "{address}", output.Address().Bech32(deps.Bech32HRP))
-	addressBech32TopicHasSubscribers := deps.MQTTBroker.HasSubscribers(addressBech32Topic)
-
-	addressEd25519Topic := strings.ReplaceAll(topicAddressesEd25519Output, "{address}", output.Address().String())
-	addressEd25519TopicHasSubscribers := deps.MQTTBroker.HasSubscribers(addressEd25519Topic)
+	//addressBech32Topic := strings.ReplaceAll(topicAddressesOutput, "{address}", output.Address().Bech32(deps.Bech32HRP))
```
muXxer, line 206:
```suggestion
	// TODO: Re-Add address topics if Indexer is enabled
	//addressBech32Topic := strings.ReplaceAll(topicAddressesOutput, "{address}", output.Address().Bech32(deps.Bech32HRP))
```

plugins/faucet/plugin.go
```diff
@@ -145,7 +151,17 @@ func provide(c *dig.Container) {
...
+	if Plugin.Node.IsSkipped(restapiv2.Plugin) {
+		Plugin.LogPanic("RestAPIV2 plugin needs to be enabled to use the Faucet plugin")
+	}
+
+	// check if Indexer plugin is disabled
+	if Plugin.Node.IsSkipped(indexerPlugin.Plugin) {
+		Plugin.LogPanic("Indexer plugin needs to be enabled to use the Faucet plugin")
+	}
+
+	restapiv2.AddPlugin("faucet")
```
muXxer:
```suggestion
	restapiv2.AddPlugin("faucet/v1")
```

plugins/debug/plugin.go
```diff
@@ -97,7 +97,7 @@ func configure() {
 	if Plugin.Node.IsSkipped(restapi.Plugin) {
 		Plugin.LogPanic("RestAPI plugin needs to be enabled to use the Debug plugin")
 	}
-	restapiv1.AddFeature(Plugin.Name)
+	restapiv2.AddPlugin("debug")
```
muXxer:
```suggestion
	restapiv2.AddPlugin("debug/v1")
```

plugins/indexer/plugin.go
```diff
@@ -0,0 +1,180 @@
...
+
+func configure() {
+	restapiv2.AddPlugin("indexer/v1")
+
+	routeGroup := deps.Echo.Group("/api/plugins/indexer/v1")
+	configureRoutes(routeGroup)
+
+	initializeIndexer()
+
+	if err := Plugin.Node.Daemon().BackgroundWorker("Close Participation database", func(ctx context.Context) {
```
muXxer:
```suggestion
	if err := Plugin.Node.Daemon().BackgroundWorker("Close Indexer database", func(ctx context.Context) {
```

plugins/indexer/plugin.go
```diff
@@ -0,0 +1,180 @@
...
+func attachEvents() {
+	deps.Tangle.Events.LedgerUpdated.Attach(onLedgerUpdated)
+}
+
+func detachEvents() {
+	deps.Tangle.Events.LedgerUpdated.Detach(onLedgerUpdated)
+}
+
+func initializeIndexer() {
+	//Compare Indexer ledgerIndex with UTXO ledgerIndex and if it does not match, drop tables and import unspent outputs
```
muXxer:
```suggestion
	// compare Indexer ledgerIndex with UTXO ledgerIndex and if it does not match, drop tables and import unspent outputs
```

plugins/debug/debug.go
```diff
@@ -122,44 +119,23 @@ func computeWhiteFlagMutations(c echo.Context) (*computeWhiteFlagMutationsRespon
...
+		return nil, err
 	}
-	return opts, nil
-}
-
-func outputsIDs(c echo.Context) (*outputIDsResponse, error) {
 
 	outputIDs := []string{}
 	outputConsumerFunc := func(output *utxo.Output) bool {
+		if filterType != nil && output.OutputType() != *filterType {
```
muXxer: Maybe refactor this like you did with the `appendConsumerFunc` in the `unspentOutputsIDs` func.

plugins/debug/debug.go
```diff
@@ -181,13 +162,19 @@ func unspentOutputsIDs(c echo.Context) (*outputIDsResponse, error) {
...
+			if output.OutputType() == *filterType {
+				return appendConsumerFunc(output)
+			}
+			return true
+		}
 	}
-	opts = append(opts, filter...)
 
 	err = deps.UTXOManager.ForEachUnspentOutput(outputConsumerFunc, opts...)
+
```
muXxer:
```suggestion
```

plugins/debug/debug.go
```diff
@@ -199,101 +186,27 @@ func unspentOutputsIDs(c echo.Context) (*outputIDsResponse, error) {
...
-	addressMap := map[string]*address{}
-
-	outputConsumerFunc := func(output *utxo.Output) bool {
-
-		if addr, exists := addressMap[output.Address().String()]; exists {
-			// add balance to total balance
-			addr.Balance += output.Amount()
+	outputIDs := []string{}
+	spentConsumerFunc := func(spent *utxo.Spent) bool {
+		if filterType != nil && spent.OutputType() != *filterType {
```
muXxer: Maybe refactor this like you did with the `appendConsumerFunc` in the `unspentOutputsIDs` func.

plugins/indexer/routes.go
```diff
@@ -0,0 +1,382 @@
...
+
+const (
+
+	// RouteOutputs is the route for getting outputs filtered by the given parameters.
+	// GET with query parameter returns all outputIDs that fit these filter criteria (query parameters: "address", "requiresDustReturn", "sender", "tag").
+	// Returns an empty list if no results are found.
+	RouteOutputs = "/outputs"
+
+	// RouteAliases is the route for getting aliases filtered by the given parameters.
+	// GET with query parameter  returns all outputIDs that fit these filter criteria (query parameters: "stateController", "governor", "issuer", "sender").
```
muXxer:
```suggestion
	// GET with query parameter returns all outputIDs that fit these filter criteria (query parameters: "stateController", "governor", "issuer", "sender").
```

plugins/indexer/routes.go
```diff
@@ -0,0 +1,382 @@
...
+	// QueryParameterAddress is used to filter for a certain address.
+	QueryParameterAddress = "address"
+
+	// QueryParameterIssuer is used to filter for a certain issuer.
+	QueryParameterIssuer = "issuer"
+
+	// QueryParameterSender is used to filter for a certain sender.
+	QueryParameterSender = "sender"
+
+	// QueryParameterTag is used to filter for a certain tags.
```
muXxer:
```suggestion
	// QueryParameterTag is used to filter for a certain tag.
```

plugins/indexer/routes.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+		filters = append(filters, indexer.ExtendedOutputSender(sender))
+	}
+
+	if len(c.QueryParam(QueryParameterTag)) > 0 {
+		indexBytes, err := restapi.ParseHexQueryParam(c, QueryParameterTag, iotago.MaxTagLength)
+		if err != nil {
+			return nil, err
+		}
+		filters = append(filters, indexer.ExtendedOutputTag(indexBytes))
```
muXxer:
```suggestion
		tagBytes, err := restapi.ParseHexQueryParam(c, QueryParameterTag, iotago.MaxTagLength)
		if err != nil {
			return nil, err
		}
		filters = append(filters, indexer.ExtendedOutputTag(tagBytes))
```

plugins/indexer/routes.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+		filters = append(filters, indexer.NFTSender(sender))
+	}
+
+	if len(c.QueryParam(QueryParameterTag)) > 0 {
+		indexBytes, err := restapi.ParseHexQueryParam(c, QueryParameterTag, iotago.MaxTagLength)
+		if err != nil {
+			return nil, err
+		}
+		filters = append(filters, indexer.NFTTag(indexBytes))
```
muXxer:
```suggestion
		tagBytes, err := restapi.ParseHexQueryParam(c, QueryParameterTag, iotago.MaxTagLength)
		if err != nil {
			return nil, err
		}
		filters = append(filters, indexer.NFTTag(tagBytes))
```

pkg/model/utxo/output.go
```diff
@@ -1,174 +1,143 @@
...
+	return o.output
 }
 
-func (o *Output) Amount() uint64 {
-	return o.amount
+func (o *Output) hasSpendingConstraint() bool {
+	switch output := o.output.(type) {
+	case iotago.UnlockConditionOutput:
+		conditions := output.UnlockConditions().MustSet()
+		return conditions.HasDustDepositReturnCondition() || conditions.HasExpirationCondition() || conditions.HasTimelockCondition()
```
muXxer: Maybe we should move that to iota.go instead?

pkg/model/utxo/output.go
```diff
@@ -1,174 +1,143 @@
...
-
-	var outputID iotago.UTXOInputID
-	copy(outputID[:iotago.TransactionIDLength], txID[:])
-	copy(outputID[iotago.TransactionIDLength:iotago.TransactionIDLength+serializer.UInt16ByteSize], bytes)
-
-	amount, err := output.Deposit()
-	if err != nil {
-		return nil, err
+	var output iotago.Output
+	if len(transaction.Essence.Outputs) < int(index) {
```
muXxer: Is that correct?
Shouldn't it be:
```go
	if len(transaction.Essence.Outputs) <= int(index) {
```

pkg/model/utxo/utxo.go
```diff
@@ -107,6 +105,11 @@ func (u *Manager) PruneMilestoneIndexWithoutLocking(msIndex milestone.Index, pru
 			mutations.Cancel()
 			return err
 		}
+
+		if err := deleteOutputLookups(spent.output, mutations); err != nil {
+			mutations.Cancel()
+			return err
+		}
```
muXxer: I don't get that part here.
The output lookups are the "unspent" lookups.... Why would we delete them?

pkg/model/utxo/snapshot.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := binary.Read(reader, binary.LittleEndian, &outputLen); err != nil {
+		return nil, fmt.Errorf("unable to read LS output length: %w", err)
+	}
+
+	outputBytes := make([]byte, outputLen)
+	if _, err := io.ReadFull(reader, outputBytes); err != nil {
+		return nil, fmt.Errorf("unable to read LS output bytes: %w", err)
+	}
+
+	output, err := iotago.OutputSelector(uint32(outputBytes[0]))
```
muXxer, line 69: You should check if outputBytes is `len > 0`

pkg/snapshot/snapshot_file.go
```diff
@@ -385,7 +322,7 @@ func StreamSnapshotDataTo(writeSeeker io.WriteSeeker, timestamp uint64, header *
 			}
 
 			outputCount++
-			outputBytes, err := output.MarshalBinary()
+			outputBytes := output.SnapshotBytes()
 			if err != nil {
 				return nil, fmt.Errorf("unable to serialize LS output #%d: %w", outputCount, err)
 			}
```
muXxer: This can be removed I guess.

plugins/faucet/plugin.go
alexsporn: Then we also need to change the plugin route to match

plugins/debug/plugin.go
alexsporn: Then we also need to change the plugin route to match

pkg/model/utxo/utxo.go
alexsporn: We are pruning the spent outputs from the db. This is just a cleanup. Normally there should be no lookups left since the output was already marked as spent, but better be thorough :-)

pkg/model/utxo/snapshot.go
alexsporn: Not needed:
```
// ReadFull reads exactly len(buf) bytes from r into buf.
// It returns the number of bytes copied and an error if fewer bytes were read.
```

alexsporn: well, theoretically you could have outputLen == 0, but then everything is wrong anyway. Let me add the check before the ReadFull

pkg/indexer/types.go
alexsporn: I created a separate ticket for this #1282

pkg/indexer/indexer.go
alexsporn: I added #1283 for this

pkg/indexer/extended.go
alexsporn: I added a ticket for this #1284

pkg/indexer/types.go
alexsporn: Added #1285

pkg/indexer/nft.go
alexsporn: See #1284

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.0.0-alpha1] - Unreleased
+
+### Changed
+    - Everything
```
muXxer:
```suggestion
    - Everything, since we are all stardust. 
```

muXxer (approved):

PR #1288, opened 2022-01-26
Add config for stardust-testnet

muXxer:


config.json
```diff
@@ -213,27 +163,14 @@
     "advancementRange": 150
   },
   "spammer": {
-    "message": "IOTA - A new dawn",
+    "message": "We are all made of stardust.",
     "index": "HORNET Spammer",
```
alexsporn: this should be tag and tagSemiLazy

docker-compose.yml
```diff
@@ -24,10 +24,6 @@ services:
...
     volumes:
       - ./config.json:/app/config.json:ro
-      - ./config_devnet.json:/app/config_devnet.json:ro
-      - ./config_comnet.json:/app/config_comnet.json:ro
       - ./peering.json:/app/peering.json
       - ./profiles.json:/app/profiles.json:ro
-      - ./mainnetdb:/app/mainnetdb
-      - ./p2pstore:/app/p2pstore
-      - ./snapshots/mainnet:/app/snapshots/mainnet
+      - ./stardust_testnet:/app/stardust_testnet
```
alexsporn, line 32: I actually don't like that the network name is hardcoded inside the docker container

config.json
muXxer: Fixed

docker-compose.yml
muXxer: I would leave it as it is for now for the alpha tests.

shufps (approved):
looks good to me :thinking:

PR #1298, opened 2022-02-02
API+Indexer refactor

alexsporn:
- Refactors the Indexer API to use an opaque cursor
- Removed `data`-envelope from API v2
- Re-structured info-endpoint and added new protocol parameters to the response

### DIFF
--- a/plugins/indexer/plugin.go
+++ b/plugins/indexer/plugin.go
@@ -113,7 +113,7 @@ func configure() {
 	}, workerpool.WorkerCount(workerCount), workerpool.QueueSize(workerQueueSize), workerpool.FlushTasksAtShutdown(true))
 
 	onLedgerUpdated = events.NewClosure(func(index milestone.Index, newOutputs utxo.Outputs, newSpents utxo.Spents) {
-		onLedgerUpdatedWorkerPool.TrySubmit(index, newOutputs, newSpents)
+		onLedgerUpdatedWorkerPool.Submit(index, newOutputs, newSpents)
 	})
 }
 
--- a/plugins/indexer/plugin.go
+++ b/plugins/indexer/plugin.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/iotaledger/hive.go/configuration"
 	"github.com/iotaledger/hive.go/events"
+	"github.com/iotaledger/hive.go/workerpool"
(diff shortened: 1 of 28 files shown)


plugins/restapi/v2/types.go
```diff
@@ -37,6 +32,28 @@ type infoResponse struct {
...
+	ReferencedRate float64 `json:"referencedRate"`
+}
+
+// infoResponse defines the response of a GET info REST API call.
+type infoResponse struct {
+	// The name of the node software.
+	Name string `json:"name"`
+	// The semver version of the node software.
+	Version string     `json:"version"`
+	Status  nodeStatus `json:"status"`
```
muXxer:
```suggestion
	// The current status of this node.
	Status  nodeStatus `json:"status"`
```

muXxer (approved):

plugins/indexer/plugin.go
```diff
@@ -102,31 +107,31 @@ func configure() {
...
 	}
 
-	configureEvents()
+	onLedgerUpdatedWorkerPool = workerpool.New(func(task workerpool.Task) {
+		updateIndexer(task.Param(0).(milestone.Index), task.Param(1).(utxo.Outputs), task.Param(2).(utxo.Spents))
+		task.Return(nil)
+	}, workerpool.WorkerCount(workerCount), workerpool.QueueSize(workerQueueSize), workerpool.FlushTasksAtShutdown(true))
+
+	onLedgerUpdated = events.NewClosure(func(index milestone.Index, newOutputs utxo.Outputs, newSpents utxo.Spents) {
+		onLedgerUpdatedWorkerPool.TrySubmit(index, newOutputs, newSpents)
```
muXxer: `TrySubmit` would mean we can loose ledger updates :scream:

muXxer (changes requested):

plugins/indexer/plugin.go
alexsporn: `workerQueueSize = 10000` not really...

muXxer: If the queue is full, the node should wait, not drop the event!

alexsporn: Ok, I'll change it to `Submit`

muXxer (approved):

PR #1300, opened 2022-02-02
Chore/cleanup tools

muXxer:


pkg/toolset/database_hash.go
```diff
@@ -122,25 +128,67 @@ func calculateDatabaseLedgerHash(dbStorage *storage.Storage) error {
...
+			Treasury:               treasury,
+			LedgerIndex:            ledgerIndex,
+			SnapshotIndex:          snapshotInfo.SnapshotIndex,
+			UTXOsCount:             len(outputs),
+			SEPsCount:              len(solidEntryPoints),
+			LedgerStateHash:        hex.EncodeToString(snapshotHashSumWithoutSEPs),
+			LedgerStateHashWithSEP: hex.EncodeToString(snapshotHashSumWithSEPs),
+		}
+
+		output, err := json.MarshalIndent(result, "", "  ")
```
alexsporn: should we refactor this into a `func printJSON(obj interface{})` and use it in every tool?

pkg/toolset/database_migration.go
```diff
@@ -8,42 +8,65 @@ import (
...
+	flag "github.com/spf13/pflag"
 
 	"github.com/gohornet/hornet/pkg/database"
 	"github.com/gohornet/hornet/pkg/utils"
 	"github.com/iotaledger/hive.go/configuration"
 	"github.com/iotaledger/hive.go/kvstore"
 )
 
+const (
+	FlagToolDatabaseMigrationDatabasePathSource   = "databasePathSource"
```
alexsporn: I like it more like it was before `sourceDatabasePath`

pkg/toolset/jwt.go
```diff
@@ -19,33 +19,31 @@ import (
...
 		return err
 	}
 
-	// Check if all parameters were parsed
-	if fs.NArg() != 0 {
-		fs.Usage()
-		os.Exit(2)
-	}
-
 	salt := nodeConfig.String(restapi.CfgRestAPIJWTAuthSalt)
```
alexsporn: do we know if this work with `--config`? Else better remove the config.json dependance altogether and let the users pass the required values as flags

pkg/toolset/snap_merge.go
```diff
@@ -2,60 +2,153 @@ package toolset
...
 
+	flag "github.com/spf13/pflag"
+
+	"github.com/gohornet/hornet/pkg/model/milestone"
 	"github.com/gohornet/hornet/pkg/snapshot"
 	"github.com/iotaledger/hive.go/configuration"
 )
 
+const (
+	FlagToolSnapshotMergeSnapshotPathTarget = "snapshotPathTarget"
```
alexsporn: `targetSnapshotPath`

alexsporn (approved):

pkg/toolset/jwt.go
muXxer: I removed it!

PR #1317, opened 2022-03-01
Stardust MQTT topics

alexsporn:


plugins/mqtt/utils.go
```diff
@@ -173,74 +191,143 @@ func publishMessageMetadata(cachedMetadata *storage.CachedMetadata) {
...
+	}
+
+	stateController := unlockConditions.StateControllerAddress()
+	if stateController != nil {
+		addr := stateController.Address.Bech32(deps.Bech32HRP)
+		publishOnTopicIfSubscribed(topicFunc(unlockConditionStateController, addr), payloadFunc())
+		addressesToPublishForAny[addr] = struct{}{}
+	}
+
+	governor := unlockConditions.StateControllerAddress()
```
muXxer: Shouldn't that be the governor?

muXxer (changes requested):

plugins/mqtt/utils.go
alexsporn: Yes! Thanks

muXxer (approved):

PR #1331, opened 2022-03-10
Feat/add database verify tool

muXxer:
This PR adds a tool to verify an existing database.
It checks if all messages are found in the database, and if the ledger state changes were correct.

### DIFF
--- a/pkg/model/utxo/milestone_diff.go
+++ b/pkg/model/utxo/milestone_diff.go
@@ -151,13 +151,17 @@ func (ms *MilestoneDiff) kvStorableLoad(utxoManager *Manager, key []byte, value 
 }
 
 func (ms *MilestoneDiff) sortedOutputs() LexicalOrderedOutputs {
-	var sortedOutputs LexicalOrderedOutputs = LexicalOrderedOutputs(ms.Outputs)
+	// do not sort in place
+	sortedOutputs := make(LexicalOrderedOutputs, len(ms.Outputs))
+	copy(sortedOutputs, LexicalOrderedOutputs(ms.Outputs))
 	sort.Sort(sortedOutputs)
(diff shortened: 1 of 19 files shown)


pkg/model/utxo/milestone_diff.go
```diff
@@ -147,6 +150,31 @@ func (ms *MilestoneDiff) kvStorableLoad(utxoManager *Manager, key []byte, value
...
+	sort.Sort(sortedSpents)
+	return sortedSpents
+}
+
+func (ms *MilestoneDiff) SHA256Sum() ([]byte, error) {
+
+	msDiffHash := sha256.New()
+
+	// compute the sha256 of the milestone diff byte representation
+	if err := binary.Write(msDiffHash, binary.LittleEndian, append(ms.kvStorableKey(), ms.kvStorableValue()...)); err != nil {
```
alexsporn: why append first? You can add first the key, then the value to the hash, is probably more memory-efficient

pkg/model/utxo/utxo.go
```diff
@@ -473,3 +475,37 @@ func (u *Manager) ComputeBalance(options ...UTXOIterateOption) (balance uint64,
...
+	if err := u.ForEachUnspentOutput(func(output *Output) bool {
+		outputs = append(outputs, output)
+		return true
+	}, ReadLockLedger(false)); err != nil {
+		return nil, err
+	}
+	sort.Sort(outputs)
+
+	for _, output := range outputs {
+		if _, err := ledgerStateHash.Write(append(output.outputID[:], output.kvStorableValue()...)); err != nil {
```
alexsporn: same here, no need to append if you are just writing the bytes into the hash

alexsporn (approved):

PR #1350, opened 2022-03-21
Participation improvements

alexsporn:
- This implementation of the Participation plugin does not calculate each address rewards on every milestone confirmation. Instead it keeps a track of each participation per address so that the rewards can be calculated on-demand for each address.
- Additionally it keeps track of the rewards generated per milestone (same approach as with voting).
- New API endpoint allow fetching all participations outputs per address

_**Note: this change is not compatible with the current events in the database. Participation should be recalculated by removing the old db.**_

### DIFF
--- a/pkg/model/participation/participation_store.go
+++ b/pkg/model/participation/participation_store.go
@@ -553,15 +553,14 @@ func (pm *ParticipationManager) rewardsForTrackedParticipation(trackedParticipat
 	return rewardsForParticipation, nil
 }
 
-func (pm *ParticipationManager) StakingRewardForAddress(eventID EventID, address iotago.Address) (uint64, error) {
+func (pm *ParticipationManager) StakingRewardForAddress(eventID EventID, address iotago.Address, msIndex milestone.Index) (uint64, error) {
 	var rewards uint64
-	confirmedMilestoneIndex := pm.syncManager.ConfirmedMilestoneIndex()
 	trackedParticipations, err := pm.ParticipationsForAddress(eventID, address)
 	if err != nil {
 		return 0, err
 	}
 	for _, trackedParticipation := range trackedParticipations {
-		amount, err := pm.rewardsForTrackedParticipation(trackedParticipation, confirmedMilestoneIndex)
+		amount, err := pm.rewardsForTrackedParticipation(trackedParticipation, msIndex)
 		if err != nil {
 			return 0, err
 		}
--- a/pkg/model/participation/participation_store.go
+++ b/pkg/model/participation/participation_store.go
@@ -20,6 +20,7 @@ var (
 	ErrInvalidPreviouslyTrackedParticipation = errors.New("a previously tracked participation changed and is now invalid")
 	ErrInvalidCurrentBallotVoteBalance       = errors.New("current ballot vote balance invalid")
 	ErrInvalidCurrentStakedAmount            = errors.New("current staked amount invalid")
+	ErrInvalidCurrentRewardsAmount           = errors.New("current rewards amount invalid")
 )
 
 // Events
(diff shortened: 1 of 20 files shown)


pkg/testsuite/test_environment.go
```diff
@@ -84,23 +84,15 @@ type TestEnvironment struct {
...
-
-	// OnNewSpent callback that will be called for each spent UTXO. This is equivalent to the tangle.NewUTXOSpent event.
-	OnNewSpent OnNewSpentFunc
-
 	// OnMilestoneConfirmed callback that will be called at confirming a milestone. This is equivalent to the tangle.MilestoneConfirmed event.
 	OnMilestoneConfirmed OnMilestoneConfirmedFunc
 
-	// OnConfirmedMilestoneIndexChanged callback that will be called after confirming a milestone. This is equivalent to the tangle.ConfirmedMilestoneIndexChanged event.
-	OnConfirmedMilestoneIndexChanged OnConfirmedMilestoneIndexChangedFunc
+	// OnLedgerUpdatedFunc callback that will be called after the ledger gets updating during confirmation. This is equivalent to the tangle.LedgerUpdated event.
```
muXxer:
```suggestion
	// OnLedgerUpdatedFunc callback that will be called after the ledger gets updated during confirmation. This is equivalent to the tangle.LedgerUpdated event.
```

pkg/model/participation/participation_store.go
```diff
@@ -145,6 +146,66 @@ func participationKeyForEventAndSpentOutputID(eventID EventID, outputID *iotago.
...
+	m := marshalutil.New(33)
+	m.WriteByte(ParticipationStoreKeyPrefixTrackedOutputByAddress) // 1 byte
+	m.WriteBytes(eventID[:])                                       // 32 bytes
+	return m.Bytes()
+}
+
+func participationKeyForEventAndAddressPrefix(eventID EventID, addressBytes []byte) []byte {
+	m := marshalutil.New(66)
+	m.WriteByte(ParticipationStoreKeyPrefixTrackedOutputByAddress) // 1 byte
+	m.WriteBytes(eventID[:])                                       // 32 bytes
```
muXxer:
```suggestion
	m.WriteBytes(participationKeyForEventPrefix(eventID)) // 33 bytes
```

pkg/model/participation/participation_store.go
```diff
@@ -491,58 +498,84 @@ func (pm *ParticipationManager) stopCountingBallotAnswers(event *Event, vote *Pa
...
-	m.WriteBytes(addressBytes)                      // 33 bytes
-	return m.Bytes()
-}
+	event := pm.Event(trackedParticipation.EventID)
+	if event == nil {
+		return 0, ErrEventNotFound
+	}
 
-func (pm *ParticipationManager) StakingRewardForAddress(eventID EventID, address iotago.Address) (uint64, error) {
+	if event.StartMilestoneIndex() >= atIndex {
```
muXxer: Here we have the `>=` but for the Participation we check only for `>`.
Is something wrong here?

alexsporn: As discussed the milestone the event starts does not generate rewards. I updated the comment to clarify.

pkg/model/participation/participation_store.go
```diff
@@ -145,6 +146,65 @@ func participationKeyForEventAndSpentOutputID(eventID EventID, outputID *iotago.
...
+	m := marshalutil.New(33)
+	m.WriteByte(ParticipationStoreKeyPrefixTrackedOutputByAddress) // 1 byte
+	m.WriteBytes(eventID[:])                                       // 32 bytes
+	return m.Bytes()
+}
+
+func participationKeyForEventAndAddressPrefix(eventID EventID, addressBytes []byte) []byte {
+	m := marshalutil.New(66)
+	m.WriteBytes(participationKeyForEventPrefix(eventID)) // 33 bytes
+	m.WriteBytes(addressBytes)                                     // 33 bytes
```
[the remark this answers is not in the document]
alexsporn: 😡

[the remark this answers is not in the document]
muXxer: :runner:

muXxer (approved):

PR #1368, opened 2022-03-24
Backport/mainnet updates

muXxer:


pkg/model/utxo/utxo.go
```diff
@@ -355,3 +357,41 @@ func (u *Manager) AddUnspentOutput(unspentOutput *Output) error {
...
+		return nil, err
+	}
+	if err := binary.Write(ledgerStateHash, binary.LittleEndian, ledgerIndex); err != nil {
+		return nil, err
+	}
+
+	// get all UTXOs and sort them by outputID
+	var outputs LexicalOrderedOutputs
+	if err := u.ForEachUnspentOutput(func(output *Output) bool {
+		outputs = append(outputs, output)
```
alexsporn, line 378: this reads all outputs into memory. With stardust this is probably a bad idea.

alexsporn (approved):

pkg/model/utxo/utxo.go
muXxer: I will fix this in another PR :+1:

alexsporn (approved):

PR #1382, opened 2022-03-29
Fix/Check for updates

muXxer:
This PR changes the logic of the update check.

We only check for versions with the same major version as the application now.
If it is self-compiled, we don't check for updates.

If it is a pre-release, we also check for pre-releases in the same major version.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,89 @@
+package version
+
+import (
+	"strings"
+
+	goversion "github.com/hashicorp/go-version"
+	"github.com/tcnksm/go-latest"
+)
+
+// fixVersion fixes broken version strings.
+func fixVersion(version string) string {
+	ver := strings.Replace(version, "v", "", 1)
+	if !strings.Contains(ver, "-rc.") {
+		ver = strings.Replace(ver, "-rc", "-rc.", 1)
+	}
+	return ver
+}
+
+// versionIsPreRelease checks if a version is a pre-release.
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"strings"
+
+	goversion "github.com/hashicorp/go-version"
+	"github.com/tcnksm/go-latest"
+)
+
+// fixVersion fixes broken version strings.
+func fixVersion(version string) string {
```
alexsporn, line 11: this also needs handling for our `-alpha`

pkg/version/version_test.go
```diff
@@ -0,0 +1,29 @@
...
+	require.False(t, filter_self_compiled("v2.0.0-alpha1")) // major - pre-release
+
+	filter_stable_version := versionFilterFunc(fixVersion("v1.1.3"))
+	require.False(t, filter_stable_version("6875ccee"))      // self-compiled
+	require.True(t, filter_stable_version("v1.2.0"))         // same major
+	require.False(t, filter_stable_version("v1.2.0-rc1"))    // same major - pre-release
+	require.False(t, filter_stable_version("v2.0.0"))        // other major
+	require.False(t, filter_stable_version("v2.0.0-alpha1")) // other major - pre-release
+
+	filter_pre_release := versionFilterFunc(fixVersion("v1.1.3-rc1"))
```
alexsporn: Add a testcase with `-alpha`

alexsporn (approved):

PR #1390, opened 2022-04-01
Fix private tangle permissions

alexsporn:


private_tangle/bootstrap.sh
```diff
@@ -1,5 +1,46 @@
...
+  exit
+fi
+
+# Cleanup if necessary
+if [ -d "privatedb" ]; then
+  ./cleanup.sh
+fi
+
+# Build latest code
+docker-compose build
```
muXxer: Please also add the build flag here.

muXxer (approved):

PR #1456, opened 2022-04-26
Updated iota.go/v3 to use the new ProtocolParameters

alexsporn:
- Updated iota.go/v3 to use the new ProtocolParameters
- Updated INX to match latest changes
- Renamed lastMilestoneId to previousMilestoneId


pkg/toolset/database.go
```diff
@@ -135,7 +135,7 @@ type StoreMessageInterface interface {
 // message +1
 func storeMessage(dbStorage StoreMessageInterface, milestoneManager *milestonemanager.MilestoneManager, msg *iotago.Message) (*storage.CachedMessage, error) {
 
-	message, err := storage.NewMessage(msg, serializer.DeSeriModePerformValidation, iotago.ZeroRentParas)
+	message, err := storage.NewMessage(msg, serializer.DeSeriModeNoValidation, nil)
```
muXxer: Please add a ToDo here that we need to reenable the validation after we came up with an idea the protocol parameters.

muXxer (approved):

pkg/toolset/toolset.go
```diff
@@ -251,3 +251,18 @@ func getGracefulStopContext() context.Context {
...
 }
+
+func ProtocolParameters(networkName string, networkPrefix iotago.NetworkPrefix) *iotago.ProtocolParameters {
+	return &iotago.ProtocolParameters{
+		Version:     2,
+		NetworkName: networkName,
+		Bech32HRP:   networkPrefix,
+		MinPowScore: 0,
+		RentStructure: iotago.RentStructure{
+			VByteCost:    500,
```
muXxer: Maybe better set everything to zero, so we can also merge databases for networks with lower values.

muXxer (approved):

PR #1466, opened 2022-04-27
Add address with `tst` HRP

Thoralf-M:
# Description

Add address with `rms` HRP because that's the default for stardust

## Type of change

- [ ] Documentation enhancement

# Checklist:


alexsporn:
The private tangle uses “tst” as prefix

Thoralf-M:
@alexsporn should I only add that then and remove the others, or have all three?

alexsporn:
Lets just add the tst one

muXxer (approved):

alexsporn (approved):

PR #1474, opened 2022-04-29
Fix/dashboard milestone topic

muXxer:


pkg/snapshot/snapshot_file.go
```diff
@@ -96,12 +96,13 @@ func (md *MilestoneDiff) TreasuryOutput() *utxo.TreasuryOutput {
 		return nil
 	}
 	to := receipt.Transaction.Output
-	msID, err := md.Milestone.ID()
+	msIDPtr, err := md.Milestone.ID()
 	if err != nil {
 		panic(err)
 	}
+	milestoneID := *msIDPtr
```
alexsporn: you don't need this here, you can use msIDPtr directly in the copy

muXxer: That was just for readability. Do you think the compiler would allocate more memory with that line?
Should I remove it?

alexsporn (approved):

PR #1481, opened 2022-05-02
Add REST + INX PoW metrics to prometheus

### LINKED ISSUE #1334: Track and export count of remote-PoWs performed
The count of remote PoWs performed by a Hornet node could become an important metric for future discussions about disabling remote PoW on all nodes.

It would be great if it could be made available externally via the Prometheus Exporter.

muXxer:


pkg/tangle/attacher.go
```diff
@@ -76,7 +77,7 @@ func (t *Tangle) MessageAttacher(opts ...MessageAttacherOption) *MessageAttacher
 	}
 }
 
-func (a *MessageAttacher) AttachMessage(ctx context.Context, msg *iotago.Message) (hornet.MessageID, error) {
+func (a *MessageAttacher) AttachMessage(ctx context.Context, msg *iotago.Message, powCompletedCallback PoWCompletedCallbackFunc) (hornet.MessageID, error) {
```
alexsporn: why don't you pass a metrics object as MessageAttacherOption instead of adding a callback here?

alexsporn (approved):

PR #1509, opened 2022-05-17
Use iotago.BlockID instead of hornet.BlockID

### LINKED ISSUE #1505: Adapt to IOTA 2.0 terminology

alexsporn:


pkg/dag/concurrent_parents_traverser.go
```diff
@@ -70,25 +71,25 @@ func NewConcurrentParentsTraverser(parentsTraverserStorage ParentsTraverserStora
...
+			var inboundQueue iotago.BlockIDs
 
-			nextValue := func() hornet.BlockID {
+			nextValue := func() iotago.BlockID {
 				// in case the inbound queue is empty, we return nil to block the nil channel
 				// produced by "outboundChannel" until the next element flows in or the
 				// inbound channel is closed.
 				if len(inboundQueue) == 0 {
-					return nil
+					return iotago.EmptyBlockID()
```
muXxer: We need to create a pointer channel here I guess.
This is golang channel magic and we need to return nil.

muXxer (approved):

PR #1555, opened 2022-06-10
Adds basic protocol manager

### LINKED ISSUE #1554: Implement basic protocol manager
The basic protocol manager:
- knows what the current protocol parameters are and the pending ones (aka updates accordingly to milestone indices)
- persists the current protocol parameters to disk
- up on initialization, reads the milestones back to below max depth to derive pending protocol parameter changes

luca-moser:
Makes the `protocfg` plugin init a `protocol.Manager` instance which keeps track of the current protocol parameters and pending ones extracted from milestones. The current protocol parameters are always persisted to disk and up on startup, either the default config protocol parameters are used or the ones available on disk.

Additionally, up on the configure stage, the `protocol.Manager` examines the milestones back `belowMaxDepth` from the current confirmed milestone to reconstruct the pending protocol changes.

`isHealthy` is now dependent on whether there is a pending protocol parameter change with a protocol version which is not supported, in which case healthy is set to false (to inform clients soon enough that the node will not function further).

The `protocol.Manager` shuts down the node if it encounters a critical error.

Since there is a cyclic dependency between the `syncmanager.SyncManager`'s requirement of knowing the below max depth parameter on `New()` and the `protocol.Manager` needing the SyncManager for the above mentioned reconstruction of pending protocol parameters, this function is only executed within `configure` of `protocfg` as to break the cyclic dependency.

All code paths or constructors which took a `iotago.ProtocolParameters` are now taking a `protocol.Manager` and use its `Current()` method to get the current protocol parameters instead.

Fixes https://github.com/iotaledger/hornet/issues/1554

### DIFF
--- a/core/protocfg/core.go
+++ b/core/protocfg/core.go
@@ -52,7 +52,7 @@ func initConfigPars(c *dig.Container) error {
 
 	type cfgDeps struct {
 		dig.In
-		Storage *storage.Storage `optional:"true"`
+		Storage *storage.Storage `optional:"true"` // optional because of entry-node mode
 	}
 
 	type cfgResult struct {
--- a/core/protocfg/core.go
+++ b/core/protocfg/core.go
@@ -52,7 +52,7 @@ func initConfigPars(c *dig.Container) error {
 
 	type cfgDeps struct {
 		dig.In
-		Storage *storage.Storage
+		Storage *storage.Storage `optional:"true"`
 	}
 
 	type cfgResult struct {
--- a/core/protocfg/core.go
+++ b/core/protocfg/core.go
@@ -2,6 +2,13 @@ package protocfg
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/iotaledger/hive.go/app/core/shutdown"
+	"github.com/iotaledger/hive.go/events"
+	"github.com/iotaledger/hornet/pkg/model/storage"
+	"github.com/iotaledger/hornet/pkg/model/syncmanager"
+	"github.com/iotaledger/hornet/pkg/protocol"
+	"github.com/iotaledger/hornet/pkg/tangle"
 
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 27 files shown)


core/protocfg/core.go
```diff
@@ -19,47 +26,62 @@ func init() {
...
+	ProtocolManager *protocol.Manager
+	SyncManager     *syncmanager.SyncManager
+	ShutdownHandler *shutdown.ShutdownHandler
+}
+
 func initConfigPars(c *dig.Container) error {
 
+	type cfgDeps struct {
+		dig.In
+		Storage *storage.Storage `optional:"true"`
```
muXxer: Why is the storage optional?
Shouldn't it always be there anyway?

pkg/protocol/protocol_manager.go
```diff
@@ -0,0 +1,231 @@
...
+	for i := belowMaxDepthTarget; i < confMsIndex; i-- {
+		if protoParasMsOpt := m.readProtocolParasFromMilestone(i); protoParasMsOpt != nil {
+			m.pending = append(m.pending, protoParasMsOpt)
+		}
+	}
+}
+
+func (m *Manager) readProtocolParasFromMilestone(index milestone.Index) *iotago.ProtocolParamsMilestoneOpt {
+	cachedMs := m.storage.CachedMilestoneByIndexOrNil(index)
+	defer cachedMs.Release(true)
```
muXxer: You can not call Release on it, in case it is nil

muXxer (changes requested):

core/protocfg/core.go
luca-moser: If run as an entry node, there is no `storage.Storage` available.

pkg/protocol/protocol_manager.go
luca-moser: fixed

muXxer (approved):

PR #1563, opened 2022-06-15
Adds tip score updates stream to INX

luca-moser:
Emitted when tips are added or removed from the tip selector pool.


plugins/inx/server_blocks.go
```diff
@@ -188,6 +195,42 @@ func (s *INXServer) ListenToReferencedBlocks(_ *inx.NoParams, srv inx.INX_Listen
 	return ctx.Err()
 }
 
+func (s *INXServer) ListenToTipScoreUpdates(_ *inx.NoParams, srv inx.INX_ListenToTipScoreUpdatesServer) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	wp := workerpool.New(func(task workerpool.Task) {
+		tip := task.Param(0).(*tipselect.Tip)
+
+		blockMeta := deps.Storage.CachedBlockMetadataOrNil(tip.BlockID)
+		defer blockMeta.Release(true) // meta -1
```
muXxer: Call the release after the nil check

plugins/inx/server_blocks.go
```diff
@@ -55,25 +72,15 @@ func INXNewBlockMetadata(blockID iotago.BlockID, metadata *storage.BlockMetadata
...
-		var shouldPromote bool
-		var shouldReattach bool
-
 		switch tipScore {
 		case tangle.TipScoreNotFound:
 			return nil, errors.WithMessage(echo.ErrInternalServerError, "tip score could not be calculated")
 		case tangle.TipScoreOCRIThresholdReached, tangle.TipScoreYCRIThresholdReached:
-			shouldPromote = true
-			shouldReattach = false
+			m.ShouldPromote = true
```
muXxer, line 83: can we leave comments here which say that promote and reattach are false?
I liked to keep the setting of the explicit value so it is easier to read.
but a comment should also be fine :)
```suggestion
			m.ShouldPromote = true
			// reattach is false
```

plugins/inx/server_blocks.go
```diff
@@ -55,25 +72,15 @@ func INXNewBlockMetadata(blockID iotago.BlockID, metadata *storage.BlockMetadata
...
 		case tangle.TipScoreBelowMaxDepth:
-			shouldPromote = false
-			shouldReattach = true
+			m.ShouldReattach = true
```
muXxer, line 87:
```suggestion
			// promote is false
			m.ShouldReattach = true
```

plugins/inx/server_blocks.go
```diff
@@ -55,25 +72,15 @@ func INXNewBlockMetadata(blockID iotago.BlockID, metadata *storage.BlockMetadata
...
 		case tangle.TipScoreHealthy:
```
muXxer, line 88:
```suggestion
		case tangle.TipScoreHealthy:
			// promote is false
			// reattach is false
```

muXxer (changes requested):

muXxer (approved):

PR #1592, opened 2022-06-27
Update the Docker example README with a bit more information

jlvandenhout:
# Description

The current README distributed with the Docker example is quite brief. While all information is readily available on the Wiki (except for how to get the JWT to get access to your API), I though we could give a bit more information to get started properly right from the example.

## Type of change

- [x] Documentation Fix

# How Has This Been Tested?

I went through all the steps myself to set up a node.

# Checklist:

### DIFF
--- a/docker-example/README.md
+++ b/docker-example/README.md
@@ -8,13 +8,6 @@ This example lets you run an [IOTA Hornet node](https://wiki.iota.org/hornet/wel
 ## Usage
 *NOTE: The example commands assume you are using Linux.*
 
-### Download and unpack the archive
-Change to the directory where you want to run your node from. Download the version of the example release archive that you want to use, unpack it and remove the archive.
-
-```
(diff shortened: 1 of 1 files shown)


docker-example/README.md
```diff
@@ -10,22 +28,48 @@ DASHBOARD_PASSWORD=0000000000000000000000000000000000000000000000000000000000000
...
+```
+./prepare_docker.sh
+```
 
-Change the values accordingly, then run `docker-compose up -d`
+Generate a password hash and salt for the password you want to use to access your dashboard:
 
-You can check the logs using `docker-compose logs`
+```
+docker compose run hornet tool pwd-hash
```
svenger87, line 33: docker-compose

muXxer: In the latest docker versions, docker compose is integrated and then you can use `docker compose`

svenger87: > In the latest docker versions, docker compose is integrated and then you can use `docker compose`

Interesting. A fresh ubuntu 20.04 doesn't have that functionality yet.

centercirclesolutions: https://docs.docker.com/engine/install/ubuntu/

```
curl -fsSL https://get.docker.com -o get-docker.sh
DRY_RUN=1 sh ./get-docker.sh
```

muXxer (approved):

PR #1596, opened 2022-06-28
Move test util functions to tpkg

muXxer:
We have several occurrences of test functions in our code that do exactly the same.
This PR removes duplicated code by moving the test utility functions to a new package called `tpkg` (like in iota.go).

To be merged after #1593 

Attention: Set the base branch to develop before merging!


pkg/testsuite/utils.go
```diff
@@ -105,8 +105,7 @@ func (b *BlockBuilder) fromWalletOutputs() ([]*utxo.Output, uint64) {
 	} else {
 		if b.fakeInputs {
 			// Add a fake output with enough balance to create a valid transaction
-			fakeInputID := iotago.OutputID{}
-			copy(fakeInputID[:], randBytes(iotago.TransactionIDLength))
+			fakeInputID := tpkg.RandOutputID()
```
alexsporn, line 108: Better duplicate and create a RandTransactionID, because if anything changes then this will break

alexsporn: There is even one already `RandTransactionID`

muXxer: But `RandOutputID` is already correct in this context, no?

alexsporn: I read BlockID :weirdsmileguy:

alexsporn (approved):

PR #1602, opened 2022-07-05
Replaced milestone.Index with iotago.MilestoneIndex

alexsporn:
- Replaced milestone.Index with iotago.MilestoneIndex
- Using uint32 in WarpSync instead of int/int32 so we don't need to cast around


pkg/pruning/pruning.go
```diff
@@ -14,8 +14,8 @@ import (
 	"github.com/iotaledger/hornet/pkg/common"
 	"github.com/iotaledger/hornet/pkg/dag"
 	"github.com/iotaledger/hornet/pkg/database"
-	"github.com/iotaledger/hornet/pkg/model/milestone"
 	"github.com/iotaledger/hornet/pkg/model/storage"
+	storagepkg "github.com/iotaledger/hornet/pkg/model/storage"
```
muXxer, line 17: huh?

pkg/snapshot/snapshot.go
```diff
@@ -4,18 +4,18 @@ import (
...
-	"github.com/iotaledger/hornet/pkg/protocol"
-
 	"github.com/pkg/errors"
 
 	"github.com/iotaledger/hive.go/events"
 	"github.com/iotaledger/hive.go/logger"
 	"github.com/iotaledger/hive.go/syncutils"
 	"github.com/iotaledger/hornet/pkg/common"
-	"github.com/iotaledger/hornet/pkg/model/milestone"
 	"github.com/iotaledger/hornet/pkg/model/storage"
```
muXxer: Remove one of the storage imports.

muXxer (approved):

PR #1608, opened 2022-07-11
Block static peers in autopeering module

### LINKED ISSUE #1342: Prevent Autopeering to peer known static peers
In networks with a small amount of nodes, it can happen that nodes that are statically peered also try to peer via Autopeering what leads to a high count of log-entries like:

```
2022-03-16T12:37:41Z    INFO    Autopeering     [dropped event] disconnecting x:14626 / <peer.ID 12*8VRkEM>
2022-03-16T12:37:41Z    WARN    Autopeering     won't disconnect <peer.ID 12*8VRkEM> as its relation is not 'autopeered' but 'known'
2022-03-16T12:37:41Z    INFO    Autopeering     [outgoing peering] adding autopeering peer <peer.ID 12*pmJNnd>
2022-03-16T12:37:41Z    INFO    Autopeering     peer is statically peered already <peer.ID 12*pmJNnd>, removing from autopeering selection protocol
2022-03-16T12:37:41Z    INFO    Autopeering     [dropped event] disconnecting x:14626 / <peer.ID 12*pmJNnd>
2022-03-16T12:37:41Z    WARN    Autopeering     won't disconnect <peer.ID 12*pmJNnd> as its relation is not 'autopeered' but 'known'
2022-03-16T12:37:41Z    INFO    Autopeering     [outgoing peering] adding autopeering peer <peer.ID 12*r4Y432>
2022-03-16T12:37:41Z    INFO    Autopeering     peer is statically peered already <peer.ID 12*r4Y432>, removing from autopeering selection protocol
2022-03-16T12:37:41Z    INFO    Autopeering     [dropped event] disconnecting x:14626 / <peer.ID 12*r4Y432>
2022-03-16T12:37:41Z    WARN    Autopeering     won't disconnect <peer.ID 12*r4Y432> as its relation is not 'autopeered' but 'known'
2022-03-16T12:37:41Z    INFO    Autopeering     [outgoing peering] adding autopeering peer <peer.ID 12*8VRkEM>
2022-03-16T12:37:41Z    INFO    Autopeering     peer is statically peered already <peer.ID 12*8VRkEM>, removing from autopeering selection protocol
2022-03-16T12:37:41Z    INFO    Autopeering     [dropped event] disconnecting x:14626 / <peer.ID 12*8VRkEM>
```

This log lines repeat very fast continuosly.

As example, for one of the stardust-testnet servers, the log entries for Autopeering account for 92.81% of the total logging volume. (9281 entries of 10000 lines of log within the last 8 minutes, about 20 log lines per second).

It would be great if already statically peered nodes could be blocklisted in Autopeering.

muXxer:
It may happen that static peers (known/unknown relation) are also found via autopeering.
If that happens, the autopeering module tries to establish another connection, which is dropped because the peer is already known.

This PR tries to avoid this problem by utilizing the blocklist of the autopeering module.
Static peers are added to the blocklist of the autopeering module, so that the selection protocol ignores connection request from these peers.

### DIFF
--- a/plugins/autopeering/plugin.go
+++ b/plugins/autopeering/plugin.go
@@ -272,9 +272,7 @@ func configureEvents() {
 		// we block peers that are connected via manual peering in the autopeering module.
 		// this ensures that no additional connections are established via autopeering.
 		switch p.Relation {
-		case p2p.PeerRelationKnown:
-			deps.AutopeeringManager.Selection().BlockNeighbor(id.ID())
-		case p2p.PeerRelationUnknown:
+		case p2p.PeerRelationKnown, p2p.PeerRelationUnknown:
 			deps.AutopeeringManager.Selection().BlockNeighbor(id.ID())
 		}
 	})
@@ -298,7 +296,7 @@ func configureEvents() {
 			return
 		}
 
-		Plugin.LogInfof("removing: %s", peerOptErr.Peer.ID.ShortString())
+		Plugin.LogDebugf("removing: %s", peerOptErr.Peer.ID.ShortString())
 		deps.AutopeeringManager.Selection().RemoveNeighbor(id.ID())
 	})
 
@@ -317,9 +315,7 @@ func configureEvents() {
 		// this ensures that no additional connections are established via autopeering.
 		// if a peer gets updated to autopeered, we need to unblock it.
 		switch p.Relation {
-		case p2p.PeerRelationKnown:
-			deps.AutopeeringManager.Selection().BlockNeighbor(id.ID())
-		case p2p.PeerRelationUnknown:
+		case p2p.PeerRelationKnown, p2p.PeerRelationUnknown:
 			deps.AutopeeringManager.Selection().BlockNeighbor(id.ID())
 		case p2p.PeerRelationAutopeered:
 			deps.AutopeeringManager.Selection().UnblockNeighbor(id.ID())
--- a/plugins/autopeering/plugin.go
+++ b/plugins/autopeering/plugin.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/network"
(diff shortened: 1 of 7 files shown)


plugins/autopeering/plugin.go
```diff
@@ -256,29 +258,79 @@ func configureEvents() {
...
+
+		id := autopeering.ConvertPeerIDToHiveIdentityOrLog(p, Plugin.LogWarnf)
+		if id == nil {
+			return
+		}
+
+		// we block peers that are connected via manual peering in the autopeering module.
+		// this ensures that no additional connections are established via autopeering.
+		switch p.Relation {
+		case p2p.PeerRelationKnown:
```
alexsporn: you can combine both cases with a `,` if you are doing the exact same thing for both

plugins/autopeering/plugin.go
```diff
@@ -256,29 +258,79 @@ func configureEvents() {
...
+
+		// if a peer is disconnected, we need to remove it from the autopeering blacklist
+		// so that former known and unknown peers can be autopeered.
+		deps.AutopeeringManager.Selection().UnblockNeighbor(id.ID())
+
+		if peerOptErr.Peer.Relation != p2p.PeerRelationAutopeered {
+			return
 		}
+
+		Plugin.LogInfof("removing: %s", peerOptErr.Peer.ID.ShortString())
```
alexsporn: debug log?

plugins/autopeering/plugin.go
```diff
@@ -256,29 +258,79 @@ func configureEvents() {
...
+		id := autopeering.ConvertPeerIDToHiveIdentityOrLog(p, Plugin.LogWarnf)
+		if id == nil {
+			return
+		}
+
+		// we block peers that are connected via manual peering in the autopeering module.
+		// this ensures that no additional connections are established via autopeering.
+		// if a peer gets updated to autopeered, we need to unblock it.
+		switch p.Relation {
+		case p2p.PeerRelationKnown:
```
alexsporn: use `,` here too

alexsporn (approved):

PR #1662, opened 2022-08-08
Add genesis addresses to snap-gen tool

muXxer:
This PR adds the possibility to pass an optional json file with genesis snapshot balances to the `snap-gen` tool.

### DIFF
--- a/pkg/toolset/snap_gen.go
+++ b/pkg/toolset/snap_gen.go
@@ -1,10 +1,13 @@
 package toolset
 
 import (
+	"encoding/binary"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"os"
 
+	"github.com/pkg/errors"
 	flag "github.com/spf13/pflag"
 
 	"github.com/iotaledger/hive.go/configuration"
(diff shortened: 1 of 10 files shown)


pkg/toolset/snap_gen.go
```diff
@@ -179,3 +305,13 @@ func snapshotGen(args []string) error {
 	fmt.Println("Snapshot creation successful!")
 	return nil
 }
+
+func TransactionIDFromIndex(index int64) iotago.TransactionID {
```
alexsporn, line 309: Why not use uint64 as param?

pkg/toolset/snap_gen.go
```diff
@@ -15,12 +18,78 @@ import (
...
+	return nil
+}
+
 func snapshotGen(args []string) error {
 
 	fs := configuration.NewUnsortedFlagSet("", flag.ContinueOnError)
 	protocolParametersPathFlag := fs.String(FlagToolProtocolParametersPath, "", "the path to the initial protocol parameters file")
 	mintAddressFlag := fs.String(FlagToolSnapGenMintAddress, "", "the initial ed25519 address all the tokens will be minted to")
 	treasuryAllocationFlag := fs.Uint64(FlagToolSnapGenTreasuryAllocation, 0, "the amount of tokens to reside within the treasury, the delta from the supply will be allocated to 'mintAddress'")
+	genesisAddressesPathFlag := fs.String(FlagToolGenesisAddressesPath, "", "the file path to the genesis addresses file (optional)")
```
alexsporn, line 92: maybe it would be good to be able to pass 2 files. Just thinking e.g. for Shimmer where we want to pass the unmodified staking rewards file and an additional file for the supply increase

alexsporn (approved):

PR #1680, opened 2022-08-15
Fix forcetypeassert linter warnings

muXxer:
This PR fixes parts of the code where type assertions are forced.
Instead of forcing them, we check if the correct type was passed.

### DIFF
--- a/pkg/protocol/gossip/msg_proc.go
+++ b/pkg/protocol/gossip/msg_proc.go
@@ -152,10 +152,23 @@ func NewMessageProcessor(
 	)
 
 	proc.wp = workerpool.New(func(task workerpool.Task) {
-		p := task.Param(0).(*Protocol)
-		data := task.Param(2).([]byte)
 
-		switch task.Param(1).(message.Type) {
+		p, ok := task.Param(0).(*Protocol)
+		if !ok {
+			panic(fmt.Sprintf("invalid type: expected *Protocol, got %T", task.Param(0)))
+		}
+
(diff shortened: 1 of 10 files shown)


pkg/protocol/gossip/msg_proc.go
```diff
@@ -152,10 +152,23 @@ func NewMessageProcessor(
 	)
 
 	proc.wp = workerpool.New(func(task workerpool.Task) {
-		p := task.Param(0).(*Protocol)
-		data := task.Param(2).([]byte)
 
-		switch task.Param(1).(message.Type) {
+		p, ok := task.Param(0).(*Protocol)
+		if !ok {
```
alexsporn, line 157: This ones should panic imho. Those are dev errors

plugins/inx/server_utxo.go
```diff
@@ -333,8 +333,21 @@ func (s *INXServer) ListenToLedgerUpdates(req *inx.MilestoneRangeRequest, srv in
 	}
 
 	sendFunc := func(task *workerpool.Task, index iotago.MilestoneIndex) error {
-		newOutputs := task.Param(1).(utxo.Outputs)
-		newSpents := task.Param(2).(utxo.Spents)
+		newOutputs, ok := task.Param(1).(utxo.Outputs)
```
oopsmonk, line 336: better to use const/enums instead of magic numbers for the Index of `task.Param`

oopsmonk (approved):

PR #1711, opened 2022-08-29
Add new config param to disable snapshot file creation. Defaults to false.

### LINKED ISSUE #1706: Add possibility to disable snapshotting
Right now, pruning is impossible without snapshotting.
We should change the logic in a way, that we can disable snapshotting completely on the nodes (should be the default setting).

alexsporn:


pkg/snapshot/snapshot.go
```diff
@@ -115,18 +115,24 @@ func NewSnapshotManager(
...
+			s.LogPanic(common.ErrSnapshotInfoNotFound)
+
+			return 0
+		}
 
-	minimumIndex := snapshotInfo.SnapshotIndex()
-	minimumIndex -= s.snapshotDepth
-	minimumIndex -= s.solidEntryPointCheckThresholdPast
+		minimumIndex = snapshotInfo.SnapshotIndex()
+		minimumIndex -= s.snapshotDepth
```
muXxer, line 132: we need to check if `minimumIndex` is bigger than `snapshotDepth`, otherwise return 0.

pkg/snapshot/snapshot.go
```diff
@@ -115,18 +115,24 @@ func NewSnapshotManager(
...
+	}
 
+	if minimumIndex >= s.solidEntryPointCheckThresholdPast {
```
muXxer: We need to check if `minimumIndex` is bigger than `solidEntryPointCheckThresholdPast`, otherwise return 0.

muXxer (approved):

PR #1800, opened 2022-12-12
add install script

jorgemmsilva:


muXxer:
thx jorge!
we decided to not add this script though :sweat_smile:

PR #1807, opened 2023-01-04
Add /metadata endpoint to /included-block

muXxer:


plugins/coreapi/component.go
```diff
@@ -59,6 +59,10 @@ const (
 	// MIMEVendorIOTASerializer => bytes.
 	RouteTransactionsIncludedBlock = "/transactions/:" + restapipkg.ParameterTransactionID + "/included-block"
 
+	// RouteTransactionsIncludedBlockMetadata is the route for getting the block metadata that was included in the ledger for a given transaction ID.
+	// GET returns block metadata (including info about "promotion/reattachment needed").
+	RouteTransactionsIncludedBlockMetadata = "/transactions/:" + restapipkg.ParameterTransactionID + "/included-block" + "/metadata"
```
alexsporn: better join the last 2 components into a single static string, else it looks awkward

alexsporn (approved):

PR #1833, opened 2023-02-02
Add proof of inclusion plugin

muXxer:
This PR adds a "poor mans" proof of inclusion based on the signed parents of a milestone.
The parents are walked message by message until the message to proof is found.
All the needed messages in the audit path are added to the proof.

The audit path might not be the shortest possible, but it works.
Will be replaced by stardust PoI anyway.


plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+		ctx,
+		deps.Storage,
+		milestoneMessageID,
+		// traversal stops if no more messages pass the given condition
+		// Caution: condition func is not in DFS order
+		func(cachedMsgMeta *storage.CachedMetadata) (bool, error) { // meta +1
+			defer cachedMsgMeta.Release(true) // meta -1
+
+			// if the message is referenced by an older milestone, there is no need to traverse its parents
+			if referenced, at := cachedMsgMeta.Metadata().ReferencedWithIndex(); !referenced || (at < msIndex) {
```
Wollac: Since we start the traverse at a milestone, I guess there cannot be any messages that are not referenced.
So we could change that to a panic that will never be raised or add a comment that we still check it.

plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+			return nil
+		}
+
+		auditPathMessages = append(auditPathMessages, cachedMsg.Message().Message())
+		auditPathMessageIDs[messageID.ToArray()] = struct{}{}
+
+		return nil
+	}
+
+	originFound := false
```
Wollac: wouldn't target found be a better name, as we have `targetMessageID`?

plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+		nil,
+		false); err != nil {
+		if errors.Is(err, common.ErrOperationAborted) {
+			return nil, errors.WithMessage(echo.ErrInternalServerError, "operation aborted")
+		} else {
+			return nil, errors.WithMessage(echo.ErrInternalServerError, err.Error())
+		}
+	}
+
+	if !milestonePathFound {
```
Wollac: With the comment from above it should be sufficient to just check the `pathHead` against `MilestoneMessageID`.

plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+				for _, parent := range cachedMsgMeta.Metadata().Parents() {
+					if _, exists := auditPathMessageIDs[parent.ToArray()]; !exists {
+						continue
+					}
+
+					parentFound = true
+
+					// delete the found parent from the map, so we don't include several paths in the proof
+					// the path may not be the shortest, but its always the same. (will be replaced by stardust PoI anyway)
+					delete(auditPathMessageIDs, parent.ToArray())
```
Wollac: The `auditPathMessageIDs` will never contain more than one entry (otherwise you had a fork in the path).
Thus, I think it would be cleaner to only store a single variable like `pathHead` and do something like:
```
for _, parent := range cachedMsgMeta.Metadata().Parents() {
  if parent == auditPathHead {
    auditPathHead = parent
    return return addMessage(cachedMsgMeta.Metadata().MessageID())
  }
}
```
This also clarifies that only one path will be created

plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	if !milestonePathFound {
+		return nil, errors.WithMessagef(echo.ErrInternalServerError, "path to milestone %d not found", msIndex)
+	}
+
+	return auditPathMessages, nil
+}
+
+func checkAuditPath(ctx context.Context, auditPathMessages []*iotago.Message, milestoneMsg *iotago.Message, message *iotago.Message) (bool, error) {
```
Wollac: In the previous function it is called `targetMessage` so consider using that name also here.

plugins/poi/poi.go
```diff
@@ -0,0 +1,248 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		m := msg
+		auditPathMessageIDs[*msgID] = m
+	}
+
+	// walk the audit path starting with the milestone
+	currentMsg := milestoneMsg
```
Wollac: This check will also pass if the `auditPathMessages` are in random order and even, under certain conditions, when they contain additional messages not lying on the path.
I guess that is fine. However, for me a "path" is really used an ordered sequence of connected messages. In that case, one could also check by walking the messages from back to front and just make sure that the predecessor is a parent of the current.

plugins/dashboard/plugin.go
```diff
@@ -171,6 +172,10 @@ func configure() {
 func run() {
 
 	e := echo.New()
+	e.Server.BaseContext = func(l net.Listener) context.Context {
```
Wollac, line 175: I have no idea what this change does and how it is connected to proof-of-inclusion. But I guess it looks good :grin:

Wollac (approved):
lgtm

alexsporn (approved):

PR #1843, opened 2023-02-23
Upgrade to latest hive.go modules

alexsporn:


core/gossip/component.go
```diff
@@ -398,42 +386,9 @@ func configureEvents() {
...
-
-	onGossipServiceProtocolTerminated = events.NewClosure(func(proto *gossip.Protocol) {
-		if proto == nil {
-			return
-		}
-
-		detachEventsProtocolMessages(proto)
-
-		// detach protocol errors
-		if proto.Events != nil && proto.Events.Errors != nil {
```
muXxer, line 411: Are we sure we don't need cleanup code for these events?

core/gossip/sting.go
```diff
@@ -3,88 +3,60 @@ package gossip
...
-	}))
-}
-
-// detachEventsProtocolMessages removes all the event handlers for sent and received messages.
-func detachEventsProtocolMessages(proto *gossip.Protocol) {
-	if proto == nil {
-		return
-	}
-
-	if proto.Parser.Events.Received != nil {
```
muXxer, line 73: Are we sure we don't need cleanup code for these events?

core/pruning/component.go
```diff
@@ -105,19 +104,16 @@ func provide(c *dig.Container) error {
...
-
-	onSnapshotHandledConfirmedMilestoneIndexChanged := events.NewClosure(func(confirmedMilestoneIndex iotago.MilestoneIndex) {
-		deps.PruningManager.HandleNewConfirmedMilestoneEvent(CoreComponent.Daemon().ContextStopped(), confirmedMilestoneIndex)
-	})
-
 	if err := CoreComponent.Daemon().BackgroundWorker("Pruning", func(ctx context.Context) {
 		CoreComponent.LogInfo("Starting pruning background worker ... done")
-		deps.SnapshotManager.Events.HandledConfirmedMilestoneIndexChanged.Hook(onSnapshotHandledConfirmedMilestoneIndexChanged)
+		unhookEvent := deps.SnapshotManager.Events.HandledConfirmedMilestoneIndexChanged.Hook(func(confirmedMilestoneIndex iotago.MilestoneIndex) {
+			deps.PruningManager.HandleNewConfirmedMilestoneEvent(CoreComponent.Daemon().ContextStopped(), confirmedMilestoneIndex)
```
muXxer: now we could even use the `ctx` of the backgroundworker, which is cleaner.

core/tangle/component.go
```diff
@@ -219,18 +215,16 @@ Please restart HORNET with one of the following flags or enable "db.autoRevalida
...
 	deps.Tangle.ConfigureTangleProcessor()
 
 	return nil
 }
 
 func run() error {
-
 	if err := CoreComponent.Daemon().BackgroundWorker("Tangle[HeartbeatEvents]", func(ctx context.Context) {
-		attachHeartbeatEvents()
+		detach := attachHeartbeatEvents()
```
muXxer: why do we call it `detach` here and in other places `unhook`? Is there a difference?

pkg/protocol/gossip/msg_proc.go
```diff
@@ -151,34 +141,6 @@ func NewMessageProcessor(
...
-
-		switch msgType {
-		case MessageTypeBlock:
-			proc.processBlockData(p, data)
-		case MessageTypeBlockRequest:
-			proc.processBlockRequest(p, data)
-		case MessageTypeMilestoneRequest:
-			proc.processMilestoneRequest(p, data)
-		}
-	}, workerpool.WorkerCount(WorkerCount), workerpool.QueueSize(WorkerQueueSize))
```
muXxer, line 180: Should we add a maximum size to the new workerpool implementations? Or at least create an issue?

pkg/tangle/tangle.go
```diff
@@ -153,32 +150,12 @@ func New(
 		receiveBlockWorkerCount:          2 * runtime.NumCPU(),
 		receiveBlockQueueSize:            10000,
 		futureConeSolidifierWorkerCount:  1, // must be one, so there are no parallel solidifications of the same cone
-		futureConeSolidifierQueueSize:    10000,
 		processValidMilestoneWorkerCount: 1, // must be one, so there are no parallel validations
-		processValidMilestoneQueueSize:   1000,
 		milestoneSolidifierWorkerCount:   2, // must be two, so a new request can abort another, in case it is an older milestone
```
muXxer, line 154: Hopefully it is no problem for the solidifier logic if we don't have the workerpool with exactly the 2 possible "slots" and not use the "TrySubmit" (which would drop more than 2 pending requests to solidify)

plugins/inx/server.go
```diff
@@ -11,16 +11,16 @@ import (
...
+	"github.com/iotaledger/hive.go/runtime/event"
+	"github.com/iotaledger/hive.go/runtime/workerpool"
 	"github.com/iotaledger/hornet/v2/pkg/common"
 	inx "github.com/iotaledger/inx/go"
 	iotago "github.com/iotaledger/iota.go/v3"
 )
 
 const (
-	workerCount     = 1
-	workerQueueSize = 10000
```
muXxer, line 23: That means that a slow client would lead to OOM error instead of blocking the node until its ready again, right?

plugins/warpsync/component.go
```diff
@@ -61,109 +52,88 @@ type dependencies struct {
...
-	deps.Tangle.Events.MilestoneSolidificationFailed.Detach(onMilestoneSolidificationFailed)
-	warpSync.Events.CheckpointUpdated.Detach(onWarpSyncCheckpointUpdated)
-	warpSync.Events.TargetUpdated.Detach(onWarpSyncTargetUpdated)
-	warpSync.Events.Start.Detach(onWarpSyncStart)
-	warpSync.Events.Done.Detach(onWarpSyncDone)
+func hookEvents() (detach func()) {
+	var heartbeakUnhook func()
+	return lo.Batch(
+		deps.GossipService.Events.ProtocolStarted.Hook(func(p *gossip.Protocol) {
+			heartbeakUnhook = p.Events.HeartbeatUpdated.Hook(func(hb *gossip.Heartbeat) {
```
muXxer: this seems wrong to me on the first look.
this heartbeat unhook is only valid for this `gossip.Protocol`, but is reused for every other `ProtocolTerminated` event.

plugins/warpsync/component.go
```diff
@@ -61,109 +52,88 @@ type dependencies struct {
...
+				warpSync.UpdateCurrentConfirmedMilestone(deps.SyncManager.ConfirmedMilestoneIndex())
+				warpSync.UpdateTargetMilestone(hb.SolidMilestoneIndex)
+			}).Unhook
+		}).Unhook,
+
+		deps.GossipService.Events.ProtocolTerminated.Hook(func(p *gossip.Protocol) {
+			if heartbeakUnhook != nil {
```
muXxer: Here, it could be the unhook for every other protocol instance.

pkg/tangle/tangle_processor.go
```diff
@@ -125,25 +52,36 @@ func (t *Tangle) RunTangleProcessor() {
 	}
 
 	if err := t.daemon.BackgroundWorker("TangleProcessor[UpdateMetrics]", func(ctx context.Context) {
-		t.Events.BPSMetricsUpdated.Hook(onBPSMetricsUpdated)
+		unhookEvents := t.Events.BPSMetricsUpdated.Hook(func(bpsMetrics *BPSMetrics) {
```
muXxer: `unhookEvents` and in other places `unhook` or `detach` :trollface:

pkg/tangle/tangle_processor.go
```diff
@@ -125,25 +52,36 @@ func (t *Tangle) RunTangleProcessor() {
...
+		unhookEvents()
 	}, daemon.PriorityMetricsUpdater); err != nil {
 		t.LogPanicf("failed to start worker: %s", err)
 	}
 
 	if err := t.daemon.BackgroundWorker("TangleProcessor[ReceiveTx]", func(ctx context.Context) {
 		t.LogInfo("Starting TangleProcessor[ReceiveTx] ... done")
-		t.messageProcessor.Events.BlockProcessed.Hook(onBlockProcessed)
-		t.Events.BlockSolid.Hook(onBlockSolid)
+		unhookEvents := lo.Batch(
```
muXxer: `unhookEvents` and in other places `unhook` or `detach` :trollface:

pkg/tangle/tangle_processor.go
```diff
@@ -164,17 +103,40 @@ func (t *Tangle) RunTangleProcessor() {
 	if err := t.daemon.BackgroundWorker("TangleProcessor[ProcessMilestone]", func(ctx context.Context) {
 		t.LogInfo("Starting TangleProcessor[ProcessMilestone] ... done")
 		t.processValidMilestoneWorkerPool.Start()
-		t.milestoneManager.Events.ReceivedValidMilestone.Hook(onReceivedValidMilestone)
-		t.Events.LatestMilestoneIndexChanged.Hook(onLatestMilestoneIndexChanged)
-		t.Events.MilestoneTimeout.Hook(onMilestoneTimeout)
+		unhookEvents := lo.Batch(
```
muXxer: `unhookEvents` and in other places `unhook` or `detach` :trollface:

pkg/tangle/tangle_processor.go
```diff
@@ -261,7 +224,12 @@ func (t *Tangle) processIncomingTx(incomingBlock *storage.Block, requests gossip
 
 				if isSolid {
 					// try to solidify the future cone of the block
-					t.futureConeSolidifierWorkerPool.Submit(cachedBlock.CachedMetadata()) // meta pass +1
+					metadata := cachedBlock.CachedMetadata() // meta +1
```
muXxer: Better call it `cachedMeta`

pkg/tangle/tangle_processor.go
```diff
@@ -290,28 +258,20 @@ func (t *Tangle) processIncomingTx(incomingBlock *storage.Block, requests gossip
...
-
-// DeregisterBlockProcessedEvent removes a registered event to free the memory if not used.
-func (t *Tangle) DeregisterBlockProcessedEvent(blockID iotago.BlockID) {
-	t.blockProcessedSyncEvent.DeregisterEvent(blockID)
-}
-
-// RegisterBlockSolidEvent returns a channel that gets closed when the block is marked as solid.
-func (t *Tangle) RegisterBlockSolidEvent(blockID iotago.BlockID) chan struct{} {
-	return t.blockSolidSyncEvent.RegisterEvent(blockID)
+// BlockProcessedListener returns a listner that can be used to wait until the block is processed.
```
muXxer:
```suggestion
// BlockProcessedListener returns a listener that can be used to wait until the block is processed.
```

muXxer (changes requested):

plugins/inx/server.go
alexsporn: Yes

muXxer (approved):

PR #1899, opened 2023-08-02
Add REST API and JWT

muXxer:


plugins/webapi/balances.go
```diff
@@ -1,82 +1,88 @@
...
 		balance, _, err := tangle.GetBalanceForAddressWithoutLocking(hornet.HashFromAddressTrytes(addr))
 		if err != nil {
-			e.Error = "Ledger state invalid"
-			c.JSON(http.StatusInternalServerError, e)
-			return
+			return nil, errors.WithMessage(echo.ErrInternalServerError, err.Error())
 		}
 
 		// Address balance
-		result.Balances = append(result.Balances, strconv.FormatUint(balance, 10))
```
shufps: should it be string?

